PR #39, opened 2015-05-26
docs for system interrupt handlers

m-mcgowan:
creating a PR for peer review.

### DIFF
--- a/src/content/en/shared/firmware.md
+++ b/src/content/en/shared/firmware.md
@@ -5120,3 +5120,63 @@ for (int i = 0; i < arraySize(myPins); i++) {
   Serial.println(myPins[i]);
 }
 ```
+
+## System Interrupts
+
+Since v0.4.2. This is advanced, low-level functionality, intended primarily for library writers.
+
+System interrupts happen as a result of peripheral events within the system. These
+system interrupts are supported on all platforms:
+
+Identifier | Description |
+-----------------------------------------
+SysInterrupt_SysTick | System Tick (1ms) handler |
+SysInterrupt_TIM3 | Timer 3 interrupt | 
+SysInterrupt_TIM4 | Timer 4 interrupt |
+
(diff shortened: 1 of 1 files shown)


andyw-lala:
I think you should explicitly state whether or not the tone/pwm timers are started and active regardless of whether any tone or pwm functions are used, as well as the frequency at which the interrupts occur.

Also specify the relationship between other events resulting from the interrupt (e.g. do these hooks get called before/after other code.)

jme783:
@m-mcgowan assuming that this was added to our new firmware page in the new docs. If not, feel free to re-open the P/R. Closing for now

technobly:
Yep, we need this in the docs :)

technobly:
@m-mcgowan I was looking to convert this over to the new Docs (currently this is written for the old Docs)... and I wanted to just get the info out there so people can start consuming it.  However, after reading it I'm not exactly sure how these system interrupts are initialized or configured.  The handlers make sense, but what "fires" these?  A concrete example with a timer with values would go a long way.  Thanks!

pkourany:
Folks, there is a much longer list of timer interrupts than the ones you list.  Each timer also has 4 channels, each with interrupts for example.  I believe a more involved explanation may be warranted.

technobly:
You know it @pkourany!  Hey are you using this for your TimerInterval library?  Any chance you want to add some details?  I know writing docs can be a bore, but when you are really into a feature the details can flow pretty easily.  Any help would be appreciated, even if you just paste in a simple example.

ensonic:
I am a bit late here, but can't figure how to make regullar irq handlers like:
'''
void DMA1_Stream5_IRQHandler(void)
'''

If you only support a subset (like timers), can you please explicitely say that?

chobberoni:
@m-mcgowan what's the status on this PR? merge of too old of a format to merge?

m-mcgowan:
I'll submit a new PR after copying the changes into the current docs.

PR #59, opened 2015-06-06
Changed "install" to "uninstall" for spark-cli

don-smith:
I think this is what you meant, no?


kennethlimcp:
Definitely! @locksmithdon can you also change that for the **core** so that i can pull it in?

Git blame by @jme783 : https://github.com/spark/docs/commit/1d37ad0521a1f524364139eae5f7c28cbb0a9545

PR #69, opened 2015-06-12
Change 'event' to 'eventName' in JSON boilerplate

starsimpson:
Found that 'event' should be 'eventName' in JSON boilerplate for e.g. integrating with Librato. I believe this should be reflected throughout the webhooks docs.. and I believe I edited all of the correct examples.

### DIFF
--- a/src/content/en/shared/webhooks.md
+++ b/src/content/en/shared/webhooks.md
@@ -66,7 +66,7 @@ Sending webhook request  { uri: 'https://api.particle.io/v1/webhooks',
   method: 'POST',
   json: true,
   form:
-   { event: 'slack-incoming-webhook',
+   { eventName: 'slack-incoming-webhook',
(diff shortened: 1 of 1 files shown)


kennethlimcp:
That's actually an issue with `particle-cli`. The fix i sent in is here: https://github.com/spark/particle-cli/pull/35

dmiddlecamp:
Hi @starsimpson,

  Thanks for the PR!  Much appreciated!  :)  Sorry about the delay evaluating this one, I wanted a chance to really thoroughly review it to be sure.

I think @kennethlimcp is right in this case, the CLI might be misleading if its saying "eventName", but the API is definitely expecting a parameter named "event".  I think it makes sense to fix the CLI to make sure it's not misrepresenting what its sending.  

If that seems wrong, could you give any more background over what wasn't working with the demos, or how you were making the requests?

jme783:
Closing this as is not a valid enhancement to the docs

PR #74, opened 2015-06-22
Describe what the mysterious signal does.

monkbroc:
Was that meant to be a secret @dmiddlecamp?

Nyan cat

### DIFF
--- a/src/content/en/shared/javascript.md
+++ b/src/content/en/shared/javascript.md
@@ -385,28 +385,28 @@ device.rename('new-name', function(err, data) {
 
 ### signal
 
-Send a signal to the device
+Send a signal to the device to start shouting rainbows like when first claiming the device. 
 
 ```javascript
 device.signal(function(err, data) {
   if (err) {
     console.log('Error sending a signal to the device:', err);
   } else {
-    console.log('Signal sent successfully:', data);
+    console.log('The LED is blinking like a rainbow:', data);
   }
 });
 ```
 
 ### stopSignal
 
-Send stop signal to the device
+Send a signal to the device to stop shouting rainbows
 
 ```javascript
 device.stopSignal(function(err, data) {
   if (err) {
     console.log('Error sending stop signal to the device:', err);
   } else {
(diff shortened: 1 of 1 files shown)


dmiddlecamp:
Yea that's an Easter egg, not meant to be documented ;)

monkbroc:
You leave just enough evidence by mentioning it in the Javascript SDK docs but not enough to explain what it does. I respect that.

moors7:
I do think this could be useful in the (web) IDE, as an "identify" button. On more than one occasion I have flashed the wrong device because I didn't know which was which. Having the ability to make it shout rainbows would be useful. Just a thought.

dmiddlecamp:
Ah, I spoke too early, the command in the CLI is an easter egg, but the api / SDKs should definitely be documented.  :)  

Here's the secret CLI command:

particle nyan 
particle nyan [on/off]
particle nyan device_id [on/off]

:)

moors7:
Should we then hide this as to not spoil it?
https://community.particle.io/t/what-does-sparkapi-prototype-signalcore-do/12741/4

dmiddlecamp:
I'm happy if people discover it, no need to hide it any more than it is already.  :)

monkbroc:
So are you OK adding the note I put here to the Javascript SDK docs, and also adding a section in the Cloud API docs that explains you can include "signal" in the device PUT call, but leaving the nyan command out of the CLI docs?

The initial text of the PR was confusing. @dmiddlecamp assumed I proposed to document the nyan CLI command where I wanted to clarify an already existing section of the Javascript SDK docs that really puzzled me.

Actually @moors7 I found out about the nyan command because I was about to implement a `particle rainbow` command in a fork of the CLI for exactly the purpose you describe: to quickly identify a particular device from a pile. I also think it would be useful to have a "Who am I?" button in the devices view of the web IDE.

dmiddlecamp:
:+1:

jme783:
@monkbroc  assuming this is now out of date or has been added to the new docs. If not, please re-open the P/R after pulling `master` and adding to the new repo structure.

PR #78, opened 2015-06-26
Added mention of firmware version targeting

suda:


jme783:
@suda can you resolve the merge conflicts before this gets merged? Also, is this on prod yet?

suda:
Updated. It isn't on prod yet but should be after today's push

src/content/guide/getting-started/build.md
```diff
@@ -169,12 +169,17 @@ Wait, what is firmware?
...
 
+
+![Firmware versions]({{assets}}/images/ide-firmware-versions.png)
+
 *Firmware* is so-called because it's harder than software and softer than hardware. Hardware is fixed during manufacturing, and doesn't change. Software can be updated anytime, so it's very flexible. Firmware is somewhere in between; hardware companies do issue firmware updates, but they tend to be very infrequent, because upgrading firmware can be difficult.
 
 In our case, because the Cores and Photons are connected to the internet, updating firmware is quite trivial; we send it over the network, and we have put in place safeguards to keep you from "bricking" your device.
 
 When you flash code onto your device, you are doing an *over-the-air firmware update*. This firmware update overwrites almost all of the software on the device; the only piece that is untouched is the bootloader, which manages the process of loading new firmware and ensures you can always update the firmware over USB or through a factory reset.
```
m-mcgowan, line 179: Just to note that this is no longer true for the Photon.

suda, line 179: And shouldn't be true for Core after 0.4.5 right? So I think we can rewrite it after the release.

PR #84, opened 2015-07-02
Clarify "user safe mode"

### LINKED ISSUE #83: Unclear description of "user safe mode"
The docs currently describe "user safe mode" with the following sentences:

"This is the most frequently mode, because your user application will not execute.  This can be useful if you have a bug in your code that is keeping you from re-flashing it over the air, or causing other hanging issues."

Is the first sentence trying to say that this mode is typically used when the user application will not execute? These sentences seem to indicate that this mode resets the device in a way that lets you always re-flash over the air (although it doesn't actually say that).

If my guess about the meaning is correct, I would reword this as something like:

"This mode will allow your device to be re-flashed over the air even if your application hangs the device or has a bug that otherwise prevents you from re-flashing."

rephorm:
Fixes #83


src/content/en/photon/start.md
```diff
@@ -136,7 +136,7 @@ The SETUP button serves two functions after the Photon has booted:
 
 The SETUP button serves four functions before the Photon has booted:
 
-- Hold down the SETUP button, tap on the RESET button and wait for *less than 3 seconds* to enter *User Safe Mode* mode, where the RGB LED will be flashing Magenta. Release the SETUP button. This is the most frequently mode, because your user application will not execute.  This can be useful if you have a bug in your code that is keeping you from re-flashing it over the air, or causing other hanging issues.
+- Hold down the SETUP button, tap on the RESET button and wait for *less than 3 seconds* to enter *User Safe Mode* mode, where the RGB LED will be flashing Magenta. Release the SETUP button. This mode places the device in a safe state where no user code is run. This can be useful if you have a bug in your code that is keeping you from re-flashing it over the air, or causing other hanging issues.
```
kennethlimcp, line 139: let's go for  `is running`. Can you also squash the commits into one once you make the change? thanks!

technobly, line 139: This text originally said: "In this mode your user application will not execute."  BLAME: https://github.com/rephorm/docs/commit/6b97043bd0ca346e3ac4ea02c704d4711816913a

kennethlimcp:
@rephorm,

can you update the commit or i will pull this in by hand instead? Thanks! :)

m-mcgowan:
:+1 to stick with "application code" rather than "user code", since application is more specific.

PR #89, opened 2015-07-17
Add notice about event name prefix filtering

jonaswisplinghoff:
I ran into this issue today. Not having used Spark.subscribe() before I thought the event names are simply equated for triggering the webhook. So by accident on of my webhooks used an event name that was the prefix of another webhooks event name, causing the first webhook to be triggered too often and send invalid data to the backend. I think a reference to the part of the docs where the prefix filtering is explained helps others not to run into this problem.


jme783:
@jonaswisplinghoff can you re-open this P/R with the new structure of our repo? Closing for now.

PR #99, opened 2015-07-27
Add community supported clients page

monkbroc:
This PR adds a "Community Supported Clients" page with a clear notice that this code is not tested by Particle. This is where open-source Ruby, Go, Python or even Java ;-) clients written by community members can be made visible to other community members.


jme783:
Nice addition! Thanks for the contribution!

PR #100, opened 2015-07-27
Document the SOS pattern

monkbroc:
When I started development for the Core, one of the issues I had was figuring out what the blinking red light pattern was. Advanced users in the forum know that it's the SOS pattern, but that's not documented.

I would have wasted less time if I just knew that blinking red means an issue in my firmware, so do a factory and try again.

### DIFF
--- a/src/content/guide/getting-started/modes.md
+++ b/src/content/guide/getting-started/modes.md
@@ -203,3 +203,18 @@ Procedure:
 The device now is in the DFU mode.
 
 {{/if}}
+
+### SOS
+
+Is your device flashing red? Oh no!
+
+A pattern of more than 10 red flashes is caused by the firmware crashing. The pattern is 3 short flashes, 3 long flashes, 3 short flashes (SOS pattern), followed by a number of flashes that depend on the error, then the SOS pattern again.
+
+{{#if photon}}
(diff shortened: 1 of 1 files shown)


jme783:
Thanks for the P/R @monkbroc!

technobly:
Thank for the PR @monkbroc... I've heard this request a few times now and thought for sure this was in the docs.  I think it was moved to support.particle.io for some reason, but should also be in the docs.
http://support.particle.io/hc/en-us/articles/204357864-Flashing-Red

Is Device Modes the best place for this, or do we need a Troubleshooting section? Oh looks like we added the Troubleshooting section back into the new docs... will need to fold this in there, under the Led Colors (Explained) section.
cc: @jme783 @cmsunu28

kennethlimcp:
:+1: to Troubleshooting

monkbroc:
Interesting. I didn't find that support article when I first ran into the issue.

Do I understand that the support.particle.io is going away, and everything will be in docs.particle.io? That's a great change. I can't say I've ever looked at support.particle.io but I read every page of docs.

Could you add a link to the bottom of the Device Modes page to the Troubleshooting -- Led Colors section? Figuring out what LED colors means has got to be the most common question so making the answer easy to find is a must.

PR #102, opened 2015-07-31
Document how to expose a C++ method to the cloud

monkbroc:
The firmware supports exposing a method on an object with
`Spark.function` by accepting a `std::function` but figuring out how to
do it requires C++ template ninja skills.

This PR adds an example on using `std::bind` to expose the method of a
C++ object through `Spark.function`.

The example was tested to work on a Photon.

### DIFF
--- a/src/content/reference/firmware.md
+++ b/src/content/reference/firmware.md
@@ -124,6 +124,8 @@ int brewCoffee(String command)
 }
 ```
 
+---
+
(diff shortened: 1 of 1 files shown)


monkbroc:
This came out of a [forum discussion](http://community.particle.io/t/red-led-flashes-when-calling-cloud-function-and-makes-photon-reboot/13999/10?u=jvanier).

Since the syntax is so obscure for C++ beginners, it should be documented.

PR #118, opened 2015-08-12
Removed exclamation point

hunterthedog:
That exclamation point was too enthusiastic. Also this is a test PR and you do not have to merge it.


hunterthedog:
Hey

I like puppies

PR #131, opened 2015-08-16
Spark.publish() example updated

jschrempp:
Spark.publish() now returns a boolean. I updated the example to be more like the example for Spark.connected


jme783:
Nice addition, thanks!

PR #135, opened 2015-08-17
Added Red SOS codes

jschrempp:
These used to be in the documentation, but somehow got dropped. I added them at the very end of this doc, where it talked about red SOS flashes.


jme783:
Awesome, merging!

PR #144, opened 2015-08-27
Add toFloat() and remove() function

itsme6582:
Add toFloat() as a String method to the documentation

### DIFF
--- a/src/content/reference/firmware.md
+++ b/src/content/reference/firmware.md
@@ -4175,6 +4175,24 @@ Parameters:
 
 Returns: The length of the String in characters.
 
+### remove()
+
+The String `remvove()` function modifies a string, in place, removing chars from the provided index to the end of the string or from the provided index to index plus count.
+
+```C++
+// SYNTAX
+string.remove(index)
+string.remove(index,count)
+```
+
+Parameters:
+
+  * string: the string which will be modified - a variable of type String
+  * index: a variable of type unsigned int
+  * count: a variable of type unsigned int
+
+Returns: None
+
 ### replace()
 
 The String `replace()` function allows you to replace all instances of a given character with another character. You can also use replace to replace substrings of a string with a different substring.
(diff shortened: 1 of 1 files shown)


kennethlimcp:
Hmm.. interesting. I wonder why has this not been in the docs for such a long time since the implementation.

itsme6582:
It's easy to miss small things like this when dealing with big projects,
especially with looming deadlines.

On Thu, Aug 27, 2015 at 1:44 AM, Kenneth Lim notifications@github.com
wrote:

> Hmm.. interesting. I wonder why has this not been in the docs for such a
> long time since the implementation.
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/spark/docs/pull/144#issuecomment-135300799.

## 

Jesse Weiland
jesse.weiland@gmail.com

kennethlimcp:
Looks good. I will try out and see if the instructions are accurate. @m-mcgowan might be faster for you to read through the additions?

m-mcgowan:
:+1:  done, get's my vote.   I can add some API tests in firmware also.

kennethlimcp:
@itsme6582 looks like this is good to merge but I would need the 2 commits squashed into one. Are you familiar with that? Otherwise i can do it :)

itsme6582:
Sure. I'll get to it sometime over Labor Day weekend.

@itsme6582 https://github.com/itsme6582 looks like this is good to merge
but I would need the 2 commits squashed into one. Are you familiar with
that? Otherwise i can do it :)

src/content/reference/firmware.md
```diff
@@ -4301,6 +4319,21 @@ Parameters:
 
 Returns: None
 
+### toFloat()
+
+Converts a valid String to a float. The input string should start with a digit. If the string contains non-digit characters, the function will stop performing the conversion. For example, the strings "123.45", "123", and "123fish" are converted to 123.45, 123.00, and 123.00 respectively. Note that "123.456" is approximated with 123.46. Note too that floats have only 6-7 decimal digits of precision and that longer strings might be truncated.
```
kennethlimcp, line 4324: @itsme6582 can we update to : 

Converts a valid String to a float. The input string should start with a digit. If the string contains non-digit characters, the function will stop performing the conversion. For example, the strings "123.45", "123", and "123fish" are converted to 123.45, 123.00, and 123.00 respectively. Note that "123.456" is approximated with 123.46. Note that a float has only 6-7 decimal digits of precision and anything behind might be truncated.

kennethlimcp:
@itsme6582 sorry for the delay. Thanks once again! :smile_cat:

PR #148, opened 2015-08-28
Update firmware.md

geoffsharris:
Added "pinMode(D2, INPUT);" to attachinterrupt code example. If the D2 pin is not declared as an input pin, nothing happens with the attachinterrupt() (at least on a core). Since Arduino code does not require this declaration, not having this in the example is a recipe for disappointment and wasted time for those that come with this assumption that the photon/core will be just like an arduino.

### DIFF
--- a/src/content/reference/firmware.md
+++ b/src/content/reference/firmware.md
@@ -3151,6 +3151,7 @@ volatile int state = LOW;
 void setup()
 {
   pinMode(ledPin, OUTPUT);
+  pinMode(D2, INPUT);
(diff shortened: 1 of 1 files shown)


kennethlimcp:
@m-mcgowan is there an underlying `attachInterrupt()` issue here?

zsup:
yep and we created an issue over there for it (https://github.com/spark/firmware/issues/589)

monkbroc:
Closing this since the docs now specify that the pins needs the correct mode (INPUT, INPUT_PULLUP, INPUT_PULLDOWN)

PR #153, opened 2015-09-02
added core metric measurement

anneris:
Added metric measurements

### DIFF
--- a/src/content/datasheets/core-datasheet.md
+++ b/src/content/datasheets/core-datasheet.md
@@ -313,22 +313,27 @@ Mechanical drawings of the Core are available [here.](https://github.com/spark/c
 <table border = '1'>
    <tr>
       <th>Parameter</th>
-      <th>Value</th>
+      <th>Value (imperial)</th>
+      <th>Value (metric)</th>
    </tr>
    <tr>
       <td>Length</td>
       <td>1.47"</td>
+      <td>37.27 mm</td>
    </tr>
    <tr>
       <td>Width</td>
       <td>0.8"</td>
+      <td>20.32 mm</td>
    </tr>
    <tr>
       <td>Height</td>
       <td>0.5"</td>
+      <td>20.32 mm</td>
    </tr>
    <tr>
       <td>Weight</td>
+      <td>0.49 ounces</td>
       <td>14 grams</td>
    </tr>
 </table>


anneris:
Mmm, the build fails on something I did not edit

`․ERROR: 404 ON http://localhost:8080/datasheets/photon-shields/ CONTENT https://github.com/spark/photon-shields/blob/master/programmer-shield LINKS TO https://github.com/spark/photon-shields/blob/master/programmer-shield`

src/content/datasheets/core-datasheet.md
```diff
@@ -313,22 +313,27 @@ Mechanical drawings of the Core are available [here.](https://github.com/spark/c
...
    <tr>
       <th>Parameter</th>
-      <th>Value</th>
+      <th>Value (imperial)</th>
+      <th>Value (metric)</th>
    </tr>
    <tr>
       <td>Length</td>
       <td>1.47"</td>
+      <td>37.27 mm</td>
```
kennethlimcp, line 322: Can we fix this to 37.34mm (from conversion --> 37.338 mm)

You can go an edit then.. `git commit add /src*` + `git commit --amend`

kennethlimcp:
@anneris no worries the build failed for something else :)

src/content/datasheets/core-datasheet.md
anneris, line 322: I took the other approach and measured my actual Core with a calliper. 
But I doubt this little difference matters.

Also, I did this edit through the edit button in the web interface, so amending something to my commit is quite troublesome (I think?)

kennethlimcp, line 322: I'm thinking that we might as well not have the PR pulled in if the change is not made since the imperial and metric does not match up to each other if you use the value based off a vernier measurement

@cmsunu28 ?

anneris, line 322: Fine for me - as long as some kind of metric measurement is added.

technobly, line 322: Please do make the conversion match what is currently there, I.e. 37.34mm. We can dig into specs to find out what the measurement should be and what the tolerance is (but I would not just blindly trust one caliper reading as the new truth ;-) ) thanks for adding!

kennethlimcp:
Fixed with this - https://github.com/spark/docs/commit/379355154e9fd808d52f3e4e1d78132dc2d7afb5

Issue #129, opened 2015-08-15
Comments about HowTo in Windows not quite correct


ScruffR:
In 
https://docs.particle.io/guide/getting-started/connect/core/#installing-node-js
https://docs.particle.io/guide/getting-started/connect/photon/#installing-node-js
the part

``` text
Restart your computer. (You can do this by mousing over the upper right hand corner of the screen, then going to Settings > Power > Restart)
```

only refers to Windows 8 (**neither** Win 7 **nor** Win 10) and seems rather superfluous, since people should know how to restart their PC.

Similarly for checking 32bit vs. 64bit.
There is an easier and more general way, by just checking if you've got a directory `Program Files (x86)` on your system drive (yes = 64bit, no = 32bit) - unless you happen to have a multiboot system with both, but then you can expect the user to know how to find out ;-)
Alternatives are <kbd>Windows</kbd>+<kbd>Break</kbd> or just refer the user to "Control Panel" and "System" - Windows users are used to these terms or have to google them. This is what most other sites do too.

I'd be happy to file a PR, if you tell me which of the options to go for.

kennethlimcp:
@ScruffR any PR for us? :)

kennethlimcp:
@ScruffR i removed the `restart computer` description. For the 32/64 bit checking, it seems ok to have it there. Thoughts?

ScruffR:
I think the restart might not be necessary, so yes this can go, but the parentheses of how to do it **has** to go, since it's just wrong for any other than Win 8.x

Good job :+1:

kennethlimcp:
@ScruffR here..mixed PR - https://github.com/spark/docs/commit/4d9d321a092199d8b90ff8e6c2bcb70bd86435ab

So win 7 info should go too? Can we replace with the correct one?

ScruffR:
I'm still not sure about the 32/64 description. I'd rather suggest people to google it or ask someone if they don't know.
e.g. I'm using this on my tablet no mouse, so all mouse actions don't fit me. 
Win 10 is different again, ...

But I'm not sure how this is seen officially ;-)

kennethlimcp:
@ScruffR  see the latest: https://github.com/spark/docs/commit/a14f6c92d6493b4399c257f27b1ac2f1e8985cef

Going to merge in few hours time after i :zzz: if there are no violent objections

PR #162, opened 2015-09-10
Do not specify latest nodejs version

kennethlimcp:
- CLI does not work with V > 0.10.x so let's not mention latest
- partial fix for https://github.com/spark/docs/issues/129 about retarting computer

Signed-off-by: Kenneth Lim kennethlimcp@gmail.com


moors7:
It worked for me on 0.12.7, but it's throwing a bunch of errors on 4.0.0 thanks to the serialport package.

kennethlimcp:
Yup but i remember seeing some failed on v0.12.7 so we might as well say 0.10 till we are 100% tested.

PR #178, opened 2015-09-15
remove SPI DMA for core

LovespaceDave:
Core has SPI.transfer(val) function, not the DMA overload - moving if block

### DIFF
--- a/src/content/reference/firmware.md
+++ b/src/content/reference/firmware.md
@@ -1954,7 +1954,6 @@ Where the parameter, `mode` can be:
  - `SPI_MODE2`
  - `SPI_MODE3`
 
-{{#if photon}}
 ### transfer()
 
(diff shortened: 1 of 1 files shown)


kennethlimcp:
@m-mcgowan looks correct?

m-mcgowan:
Yeah, that's good!

pkourany:
What are large bytes?  I suspect you meant larger number of bytes.  Also, I would not refer to TX(output) and RX(input) as these are Serial1 references.  Perhaps SEND(output) and RECEIVE(input) are better.  One thing to note is that if you expect to receive 100 bytes, you also need to sent 100 dummy bytes so this needs to be clarified.

m-mcgowan:
I must admit I didn't read the docs - Satish wrote these some time ago and they've been languishing as a closed PR. Feel free to fix if you guys are willing and able!

PR #184, opened 2015-09-17
Adding SPI pin information and mentioning SPI1 for Photon.

dougalcampbell:
Based on [a recent thread](https://community.particle.io/t/spi-on-the-photon-confused/15623), adding additional details about SPI and SPI1.

### DIFF
--- a/src/content/reference/firmware.md
+++ b/src/content/reference/firmware.md
@@ -1867,8 +1867,7 @@ SPI1.begin(D5);
 ```{{/if}}
 
 Where, the parameter `ss` is the SPI device slave-select pin to initialize.  
-If no pin is specified, the default pin is `SS (A2)`{{#if photon}} for
-`SPI`, or `SS3 (D5)` for `SPI1`{{/if}}.
+If no pin is specified, the default pin is `SS (A2)`.
 
 ### end()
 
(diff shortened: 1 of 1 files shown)


src/content/reference/firmware.md
```diff
@@ -1846,8 +1861,14 @@ Initializes the SPI bus by setting SCK, MOSI, and a user-specified slave-select
...
 ```
+{{#if photon}}```C++
+// Example of using second port on Photon, with alternate SS pin:
+SPI1.begin(D5);
+```{{/if}}
 
-Where, the parameter `ss` is the SPI device slave-select pin to initialize.  The default pin is `SS (A2)` if no pin is specified.
+Where, the parameter `ss` is the SPI device slave-select pin to initialize.  
+If no pin is specified, the default pin is `SS (A2)`{{#if photon}} for
+`SPI`, or `SS3 (D5)` for `SPI1`{{/if}}.
```
kennethlimcp: @dougalcampbell, this is incorrect until 0.4.6 is released right?

It would be better for us to fix what's the current status since D5 is not the SPI1 default as per now.

dougalcampbell: Oh crud. I thought I changed that before I committed it. I'll redo the pull request.

kennethlimcp:
@dougalcampbell can squash the commits into one?

dougalcampbell:
Squashed PR submitted.

kennethlimcp:
Thanks @dougalcampbell  :)

PR #189, opened 2015-09-23
use A0 for servo example

smramdani:
A0 doesn't work for servo because it has no PWM, so the example was misleading.
I proposed the list of pins (from Photon doc) that have PWM as comment to help people be aware about that.

Best reguards,
Hammed


kennethlimcp:
@smramdani thanks!

PR #190, opened 2015-09-23
Specify Node.js version

fhinkel:
Latest Node.js version would include 4.\* which does not work yet.


technobly:
This falls into line with your proposed change, however I'd prefer not to put these version references everywhere because we'll forget about them and they will soon be outdated. We also have this stated at the top:

```
The Particle CLI runs with Node.js. Grab version 0.10.40 from the Node.js website
```

Does this `package.json` dependency in the CLI effectively prevent a user from installing the CLI with Node.js 4.*?
https://github.com/spark/particle-cli/blob/master/package.json#L15-L17

cc: @emilyrose :)

affinity-matrix:
@technobly it does but the error can be pretty cryptic. Hopefully the issue will soon be moot anyway!

affinity-matrix:
v1.8.9 of the Particle CLI includes support for Node 4.x via an updated dependency on node-serialport (which just released v2.0).

PR #193, opened 2015-10-01
Firmware 0.4.6

m-mcgowan:


pkourany:
@m-mcgowan, for the `retained` variables, you mention that their values will be retained "as long as the device is powered".  The point of backup RAM is that it will be retained as long as VBAT power is maintained, not primary Vin power.  Thus, this statement is incorrect

`+Note that as soon as power is lost then the contents of the backup RAM are also lost - for data to be retained the device needs to be powered._ For persistent storage of data through power loss, please use the [EEPROM](#eeprom) library.`

Again, VBAT will maintain the RTC and the backup RAM.  VBAT is the key, not Vin power.  If VBAT drops, THEN backup RAM will be lost.

So in the scenario you describe, with power being removed and VBAT kicking in, when power returns, all the retained vars will be initialized again.  This is not the intended use.  IMO, retained vars should be treated like eeprom and it is up to the user to decide when they should be initialized using a crc or other mechanism.  Am I missing something?

m-mcgowan:
The variables are retained through a software reset and sleep so long as there is power. At least, that's what I was seeing when connected to 5V USB.

pkourany:
@m-mcgowan, of course they will be retained.  The trick is when you remove power while VBAT is still being supplied.  THAT is the whole purpose of backup RAM.

m-mcgowan:
Since they are retained if you maintain either VBAT or VIN why do we need to distinguish? That's why I say maintain power - it doesn't matter which in is powered so long as one of them is. Then the data is retained.

pkourany:
From the docs it sounds that if VIN is dropped but VBAT is not, when power returns to VIN, it will appear as a full reset.  Will the backup RAM be iniatialized then?

m-mcgowan:
The backup ram will be retained - the device is powered from VBAT.  Like I said - the device retains the variables so long as power is maintained, though sleep and reset regardless of whether it's powered with VIN or VBAT.  And that is a distinction with backup RAM - the same is not true for the regular RAM which is overwritten always on reset even if power is maintained. 

Adding an extra distinction about VIN vs VBAT is IMHO an unnecessary complication - the variables are retained so long as there is power.

andyw-lala:
That's because you already understand the purpose of Vbat.

I think the distinction is helpful for people with different amounts of
background knowledge.

On Thu, Oct 1, 2015 at 10:30 AM, Matthew McGowan notifications@github.com
wrote:

> The backup ram will be retained - the device is powered from VBAT. Like I
> said - the device retains the variables so long as power is maintained,
> though sleep and reset.
> 
> Adding an extra distinction about VIN vs VBAT is IMHO an unnecessary
> complication.
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/spark/docs/pull/193#issuecomment-144763395.

## 

Andy

pkourany:
I agree.  VBAT is NOT processor power since it only powers the RTC and backup RAM.  The processor doesn't run.  VIN actually powers the processor and code runs.  I believe the distinction is important.

m-mcgowan:
But you're making a distinction above and beyond what is being discussed here. If someone want's to learn about VBAT or VIN and the different processor modes then they should check the datasheet for that. Here there is no distinction - so long as one of them is powered then the variables are retained.  I'm not discussing the other affects of the choice of using VIN or VBAT since that's outside the scope of retained variables.

andyw-lala:
You're kidding, right ?

Retained variables and vbat are inseparable.

If it wasn't for vbat there would be no retaining to talk about.

You may be trying to stake out some semantic high ground here, but we are
talking about arduino sketches here, not comp.sci research.

pkourany:
I believe the distinction is the "powered" statement.  The processor is powered by VIN ONLY.  Using VBAT is akin to a backup battery on your clock or PC motherboard.  It does NOT power the processor.  It only maintains the operation of certain elements, mainly the RTC and the backup RAM on the STM32.  The distinction is critical to the understanding of why backup RAM is needed.

m-mcgowan:
Andy, no I'm not kidding. Retained variables are also retained when VIN is maintained - it's not exclusively VBAT. They are retained through resets and sleep when either VIN or VBAT is present. Hence there is no reason to distinguish VBAT and VIN - they both make variables retained so long as either of those lines has power. I'm not trying to take some higher semantic ground, just not getting into details that don't pertain to the situation. I don't think there's more to be gained from discussing this - you're welcome to submit a PR that fixes the issue as you see it.

andyw-lala:
Everything is maintained when Vin is maintained - that's just the
definition of a running system.

I've seen tons of confusion over vbat in the user community, and am trying
to be an inside voice for those users.

If I have time, I'll write something - but this isn't actually my day job.

On Fri, Oct 2, 2015 at 3:05 PM, Matthew McGowan notifications@github.com
wrote:

> Andy, no I'm not kidding. Retained variables are also retained when VIN is
> maintained - it's not exclusively VBAT. So they are retained through resets
> and sleep when either VIN or VBAT is present. Hence there is no reason to
> distinguish VBAT and VIN - they both make variables retained so long as
> they have power. I'm not trying to take some higher semantic ground, just
> not getting into details that don't pertain to the situation. I'm done here
> - you're welcome to submit a PR that fixes the issue as you see it.
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/spark/docs/pull/193#issuecomment-145142124.

## 

Andy

bkobkobko:
Hi @m-mcgowan 

I just wanted to chime in that I am with Andy and Paul here: it doesn't read right to me.  "Powered" is an ambiguous word (or we would not be having this conversation) but generally is taken to mean "is supplied sufficient power for full operation."

If you wrote, "the last value set on the variable is retained _as long as the device is powered by either directly by USB/Vin/3.3V or in data retention mode by Vbat_. When all power is removed..." that would go a long way for me.

m-mcgowan:
> Everything is maintained when Vin is maintained - that's just the
> definition of a running system.

I think there is the mis-understanding, and the key distinction in my mind. When the system sleeps or is reset, regular variables _aren't_ retained, even when the system is fully powered. While the variables declared as `retained` are retained through reset and sleep.

zsup:
hey guys, I'm working on a PR to address the issue, coming in just a minute or two

bkobkobko:
Glad to have some editorial love, @zsup !

@m-mcgowan I don't think there is any misunderstanding on that.  It is fully clear to me and I would guess to Andy and Paul too.

zsup:
Patch attempted over on https://github.com/spark/docs/pull/195

PR #194, opened 2015-10-02
Update cli.md

dclobato:
Keeping clear that npm {install|uninstall} should be run as root

### DIFF
--- a/src/content/guide/tools-and-features/cli.md
+++ b/src/content/guide/tools-and-features/cli.md
@@ -15,7 +15,7 @@ The Particle CLI is a powerful tool for interacting with your devices and the Pa
 
   First, make sure you have [node.js](http://nodejs.org/) installed!
 
-  Next, open a command prompt or terminal, and install by typing:
+  Next, open a command prompt or terminal, and install by typing, as root or using sudo:
 
 ```sh
 # how to install the particle-cli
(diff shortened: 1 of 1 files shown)


technobly:
Thanks @dclobato, this is a persistent issue with many people.  `sudo` is the quick fix, but this is likely just a NPM permissions issue that can be solved thusly: https://docs.npmjs.com/getting-started/fixing-npm-permissions

Easy to recommend `sudo` but let's let @emilyrose comment on this. Emily, do you have a preference for instructions here?

dclobato:
Hi. Reading the docs pointed out, I do not think that change the ownership of NPM's directory is a global solution, specially on a multiuser environment... :)

technobly:
@dclobato in that case use Option 2 so you have your own NPM default directory :) https://docs.npmjs.com/getting-started/fixing-npm-permissions#option-2-change-npm-s-default-directory-to-another-directory

affinity-matrix:
Option 2 is the method I recommend to folks facing NPM permissions issues with global installs!

technobly:
:+1:

affinity-matrix:
@dclobato while this particular change won't land, I would be more than happy if you wanted to be the one to submit the patch mentioning the NPM documentation on permissions! If not I totally understand :smile_cat:

affinity-matrix:
If so, just re-open this PR with the changes, or open a new one! Thanks!

PR #199, opened 2015-10-06
Updated Node.js reference

dougseven:
I updated the Node.js reference to v0.12.7 and updated the link to the previous downloads page on NodeJs.org. I also added a note about the Windows 10 Device Manager info - on Windows 10 the Particle device shows up under Ports, not Other Devices.


kennethlimcp:
CLI now works with v4.0.0 so closing this :)

PR #204, opened 2015-10-08
String format static method should be called with colons

### LINKED ISSUE #203: String.format() example not working
I copied and paste the example from: https://docs.particle.io/reference/firmware/photon/#format- and it failed to compile for me

monkbroc:
Fixes #203


monkbroc:
Already fixed in a85cbb36512e652863392aba13aed8900c7bdd30

PR #217, opened 2015-10-21
Fix typo in example for System.version()

MisterNetwork:
Was:
// System versio: 0.4.7

Corrected:
// System version: 0.4.7


kennethlimcp:
@MisterNetwork thanks!

PR #226, opened 2015-11-14
EEPROM.write() is void

monkbroc:
EEPROM.write() doesn't return anything on error since it is void.

https://github.com/spark/firmware/blob/91172bb7864ba6580654bdd8154bcb5f1953edfc/wiring/inc/spark_wiring_eeprom.h#L128

### DIFF
--- a/src/content/reference/firmware.md
+++ b/src/content/reference/firmware.md
@@ -4507,9 +4507,6 @@ Write a byte of data to the emulated EEPROM.
 
 - On the Core, this must be a value between 0 and 99
 - On the Photon, this must be a value between 0 and 2047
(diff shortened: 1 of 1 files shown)


kennethlimcp:
@technobly

monkbroc:
Please merge a one liner docs change?

kennethlimcp:
@monkbroc, in future, referencing the firmware would help speed up the merge.

As per: https://github.com/spark/firmware/blob/develop/wiring/inc/spark_wiring_eeprom.h#L128

monkbroc:
Ok. Thanks Kenneth

PR #236, opened 2015-11-25
add comment on max I2C clock speed.

peterqu:
STM32F1xx and STM32F2xx in Core and Photon supports I2C at Standard or
Fast Mode with a max clock frequency of 400KHz.


kennethlimcp:
Pulling this in since https://github.com/spark/firmware/issues/494 has been discussed and concluded

PR #238, opened 2015-11-26
Fix IFTTT guide headers

embedded-creations:
Whitespace matters with Markdown :-)


kennethlimcp:
@embedded-creations  thanks :)

PR #241, opened 2015-11-29
Correct Subscriber in firmware.md

Kielkopf:
The Example Subscriber did not work as expected (nothing else needed ...) because the Construktor was coded as a method instead.


kennethlimcp:
@ScruffR this looks ok to me. What about you?

ScruffR:
@kennethlimcp, I'm not too sure about it, due to the unpredictability of the call sequence of constructors but the need for `Particle` to be instanciated already.

But maybe one of the C++ Jedi might have a word to add

PR #250, opened 2015-12-15
Document UDP.isValid() / UDP.status()

aledand:
spark/firmware#766 PR companion

### DIFF
--- a/src/content/reference/firmware.md
+++ b/src/content/reference/firmware.md
@@ -3262,7 +3262,7 @@ udp.leaveMulticast();
 
 ### isValid()
 
-UDP reading APIs do not check for socket validity. On typical listening loop isValid() can be used to avoid waiting for ever for data no more coming.
+UDP reading APIs do not check for socket validity. On typical listening loop isValid() can be used to avoid waiting for ever for data will never come.
 
 ```cpp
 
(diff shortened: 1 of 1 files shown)


kennethlimcp:
Ref: https://github.com/spark/firmware/pull/766

kennethlimcp:
Is this only valid for Photon?

We might need to extend the example a little and also give a description of what does function actually do...

src/content/reference/firmware.md
```diff
@@ -3260,6 +3260,22 @@ UDP udp;
 udp.leaveMulticast();
 ```
 
+### isValid()
+
+Check if the previusly bind UDP socket is valid.
```
kennethlimcp: `previously`

aledand: it is in the base code and I suppose shuld work even for Core and Electron. no way to test.

going to extend example and description ... which will be much way longer that the change itself :-)

technobly:
This may change to UDP::status() to match the existing TCPClient::status()

ScruffR:
This PR is lingering for quite some time although on the Prioritized list - why?

rickkas7:
This appears to be awaiting a code change that has not been merged in, and might also require some reworking to make the API more like TCPClient.
https://github.com/particle-iot/firmware/pull/766

PR #259, opened 2016-01-11
rename #particle-connect #particle-connect-

embedded-creations:
broken in-page links


kennethlimcp:
will need to look into this. Some links seems to have the `-` and some not.

PR #269, opened 2016-01-22
Feature/firmware 0.4.9

m-mcgowan:
docs changes for the new features in 0.4.9 firmware

### DIFF
--- a/src/content/reference/firmware.md
+++ b/src/content/reference/firmware.md
@@ -1847,14 +1847,14 @@ If `blockOnOverrun(false)` has been called, the method returns the number of byt
 
 _Since 0.4.9. Available on Serial1, Serial2, etc.. This function will support USB Serial with the next release._
 
-Defines what should happen when calls to `write()` that would overflow the buffer.  
+Defines what should happen when calls to `write()/print()/println()/printlnf()` that would overrun the buffer.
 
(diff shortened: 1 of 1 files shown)


src/content/reference/firmware.md
```diff
@@ -1258,13 +1304,17 @@ void loop()
 }
 ```
 
-- On the Core, this function works on pins A0, A1, A4, A5, A6, A7, D0 and D1.
+- On the Core, this function works on pins D0, D1, A0, A1, A4, A5, A6, A7, RX and TX.
 - On the Photon, this function works on pins D0, D1, D2, D3, A4, A5, WKP, RX and TX with a caveat: PWM timer peripheral is duplicated on two pins (A5/D2) and (A4/D3) for 7 total independent PWM outputs. For example: PWM may be used on A5 while D2 is used as a GPIO, or D2 as a PWM while A5 is used as an analog input. However A5 and D2 cannot be used as independently controlled PWM outputs at the same time.
 
-When used with these pins, the `analogWrite()` function has nothing to do with the analog pins or the `analogRead()` function.
-
+The PWM frequency must be the same for pins in the same timer group.
```
technobly: This is only true for the same timer and channel, not necessarily the overall timer regardless of channel.

PR #276, opened 2016-01-26
Add device animation for the electron

monkbroc:
@technobly Please add back the rotated and baked photon.svg, core.svg and electron.svg after merging this.


technobly:
Thanks!

PR #277, opened 2016-01-28
Update version nums for Android SDK & setup lib

jensck:


jensck:
(submitting new PR with newer versions)

PR #281, opened 2016-02-01
Adds data format examples for callFunction and getVariable for js lib

lourd:


lourd:
Bump :bug:

I also just noticed a discrepancy between the data coming in on the event stream (`getEventStream`) and the return value of `device.getVariable`. The event stream data has the field `coreid` while the getVariable has the field `coreInfo.deviceID`. Correction added to this PR

kennethlimcp:
@dmiddlecamp

dmiddlecamp:
thanks for the ping, I'll double check this and merge in when I can

PR #290, opened 2016-02-17
Particle.connect() appears to be non-blocking

Whitey2004:
Using current firmware (0.4.9) with a core dev board in semi-auto mode; calling particle.connect() begins wifi negotiation but user loop() function continues to be run even before connection is complete (still blinking green).

### DIFF
--- a/src/content/reference/firmware.md
+++ b/src/content/reference/firmware.md
@@ -411,8 +411,6 @@ void loop() {
 }
 ```
 
-After you call `Particle.connect()`, your loop will not be called again until the device finishes connecting to the Cloud. Typically, you can expect a delay of approximately one second.
-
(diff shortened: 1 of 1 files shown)


kennethlimcp:
This doesn't work for me:

```
SYSTEM_MODE(SEMI_AUTOMATIC);

void setup(){
    pinMode(D7,OUTPUT);
    digitalWrite(D7,HIGH);
    Particle.connect();
}

void loop(){
    digitalWrite(D7,LOW);
    delay(300);
    digitalWrite(D7,HIGH);
    delay(300);
}
```

But you are talking about a core.. and i tested on a Photon so not sure if the behavior is differing.

Ok indeed... Core and Photon has a different behavior...

technobly:
See issue #367

PR #292, opened 2016-02-18
Update examples.md

TechnoMaster:
Your configuration file was backwards, the code says to place your hand in front of the beam, but then we measure the on values, next we are told to remove our hand, yet we measure the off values. I switched the on and off values so we measure off when the hand is in front and on when the hand is removed.

### DIFF
--- a/src/content/guide/getting-started/examples.md
+++ b/src/content/guide/getting-started/examples.md
@@ -786,10 +786,10 @@ void setup() {
   delay(500);
 
   // Now we'll take some readings...
-  int on_1 = analogRead(photoresistor); // read photoresistor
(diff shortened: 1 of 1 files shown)


kennethlimcp:
Don't understand the issue here. 

On value is taken when your hand is blocking the photoresistor and off value is taken when your hand is not blocking the resistor. Does the example not work or something?

TechnoMaster:
The example only works because the maths to calculate the threshold can
work either way, but you have the on_1 and on_2 being calculated with the
hand blocking the photoresistor, yet you are calling the on_1 and on_2
intactValue, if my hand is blocking the photoresistor, the beam cannot be
intact

PR #294, opened 2016-02-18
WiFi.getCredentials example now compiles.

monkbroc:
- Remove fields that are always 0 from example: BSSID, RSSI and channel
- Update description to note that actual credentials are not returned

### DIFF
--- a/src/content/reference/firmware.md
+++ b/src/content/reference/firmware.md
@@ -795,7 +795,9 @@ WiFi.setCredentials("SSID", "PASSWORD", WPA2, WLAN_CIPHER_AES));
 
 *Since 0.4.9.*
 
-Lists the Wi-Fi credentials stored on the device. Returns the number of stored credentials.
+Lists the Wi-Fi networks with credentials stored on the device. Returns the number of stored networks.
+
+Note that this returns details about the Wi-Fi networks, but not the actual password.
 
 {{#if core}}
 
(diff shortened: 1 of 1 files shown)


technobly:
This caught my eye and I wondered why these fields needed to be removed for it to compile.  I found that it was because the BSSID is `uint8_t bssid[6]`.  I added a comment for WLAN_CIPHER_AES_TKIP as well.

This will compile (untested):

```
    // EXAMPLE
    WiFiAccessPoint ap[5];
    int found = WiFi.getCredentials(ap, 5);
    for (int i = 0; i < found; i++) {
        Serial.print("ssid: ");
        Serial.println(ap[i].ssid);
        Serial.printlnf("bssid: %d:%d:%d:%d:%d:%d",
            ap[i].bssid[0],ap[i].bssid[1],ap[i].bssid[2],
            ap[i].bssid[3],ap[i].bssid[4],ap[i].bssid[5]);
        // security is one of WLAN_SEC_UNSEC, WLAN_SEC_WEP, WLAN_SEC_WPA, WLAN_SEC_WPA2
        Serial.println(ap[i].security);
        // cipher is one of WLAN_CIPHER_AES, WLAN_CIPHER_TKIP, WLAN_CIPHER_AES_TKIP
        Serial.println(ap[i].cipher);
        Serial.println(ap[i].rssi);
        Serial.println(ap[i].channel);
    }
```

technobly:
Oh I see your comment now about fields that are always 0.  Yeah I'm seeing the same results, and this is only the stored network credentials, not a AP scan...

PR #296, opened 2016-02-19
added ping KB/month

jschrempp:
Everyone who reads this is going to do the calc, so why not just add it to the doc?


monkbroc:
Hi Jim, data usage is an area of the Electron that we're improving continuously. So we're not ready to cement a specific KB/month figure in people's minds for ping data usage. We'll update this section when we have updated numbers.

PR #298, opened 2016-02-21
Update firmware.md: System_thread section

juleskers:
Make more explicit that System_thread(enable) impacts more network initialisation than just Particle.\* functions
(After it stumped me for a week that UDP.joinMulticast() stopped working, see https://community.particle.io/t/system-thread-breaks-receiving-udp-multicast/20079 )

### DIFF
--- a/src/content/reference/firmware.md
+++ b/src/content/reference/firmware.md
@@ -6068,11 +6068,14 @@ When the system thread is enabled, application execution changes compared to
 non-threaded execution:
 
 - `setup()` is executed immediately regardless of the system mode, which means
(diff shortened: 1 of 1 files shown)


juleskers:
I checked the build-log, and the check seems to fail for broken links in a section I didn't touch.
Can I re-trigger the build somehow to see if was a transient failure?

kennethlimcp:
It's not an issue with the build but more of not within my jurisdiction to merge it :)

juleskers:
That's perfectly fine :-)

PR #300, opened 2016-02-21
Added DFU mode instructions for Electron

alan707:
As I was trying to setup the Electron, I noticed the instructions to put it into DFU mode were not yet written, so I've added them.


kennethlimcp:
@alan707 i changed the conditional for the existing photon instructions which is the same for the electron to prevent duplication here: https://github.com/spark/docs/commit/ddadc3257edd1d105d0ff06679be2ac4a2b024c9

Thanks!

PR #302, opened 2016-02-24
Update photon-datasheet.md

harmeetsingh-work:


kennethlimcp:
@blackadmin thanks :)

PR #303, opened 2016-02-27
Fix bad merge on key-management.md

dwaq:
It looks like a merge went bad on this file and created some leftover tags. I deleted them. I didn't run the changes locally, but it looks better in the markdown preview.


kennethlimcp:
@dwaq thanks for catching this!

PR #307, opened 2016-03-06
Fixed typo

makrandgupta:


kennethlimcp:
Sorry but this should be `a` as per the docs :/

makrandgupta:
Umm I was going through the docs and it was `an` instead of, `a`. Hence I proposed the changes

kennethlimcp:
Opps, my bad :)

Thanks!

PR #311, opened 2016-03-12
Change from "setup" button to "mode" button

MisterNetwork:
There is no button label "setup".  Button label is "MODE"

### DIFF
--- a/src/content/guide/tools-and-features/cli.md
+++ b/src/content/guide/tools-and-features/cli.md
@@ -115,7 +115,7 @@ Steps:
 # How to flash an Electron over serial
 $ particle flash --serial firmware.bin
 
-! PROTIP: Hold the SETUP button on your device until it blinks blue!
+! PROTIP: Hold the MODE button on your device until it blinks blue!
 ? Press ENTER when your device is blinking BLUE
 sending file: firmware.bin
 


kennethlimcp:
hehe.. That's tricky.... Core labels `MODE` and Photon/Electron labels `SETUP`

Maybe `MODE/SETUP`?

MisterNetwork:
Correction please - Electron label is MODE   - at least the one I received
yesterday is - which BTW now doesn't seem to be working at all.

And also the section is 
"How to flash an Electron over serial"

kennethlimcp:
@MisterNetwork , can you update to `MODE/SETUP` or i can fix this as well. Thanks! :)

technobly:
I realize this is a pain to update, but we do this all over in the docs.  Based on the Device selected in the top right of the Docs, these handlebars selectors can change the text.

Basically this say, if it's not the photon, use MODE.  If it is the photon, use SETUP.

> ! PROTIP: Hold the {{#unless photon}}MODE{{/unless}}{{#if photon}}SETUP{{/if}} button on your device until it blinks blue!

monkbroc:
How about I make a helper called `setup-button` that will insert the Right Thing :tm: in any page?

technobly:
That would be awexome @monkbroc! And it would also keep a few more fairies alive ;-) (copypasta kills fairies)

I'd call it `system-button` which is how we commonly refer to it.

technobly:
So then we'd use it like:

> ! PROTIP: Hold the {{system-button}} button on your device until it blinks blue!

monkbroc:
I implemented this change to more places in c5c54f87

PR #312, opened 2016-03-12
Removed a duplicated word

badmotorfinger:


kennethlimcp:
The shield is called a shield shield :)

technobly:
Hi @vincpa.  Thanks for the PR, but as @kennethlimcp points out this Shield is actually called a Shield Shield, because it's a Shield for Particle devices that allows you to connect other Arduino Shields to your Particle devices.  Literally a Shield for Shields.  Hope that clears up any confusion.

PR #313, opened 2016-03-15
Changed Preprocessor Pragma to newer version

mumblepins:
Newer firmware version now uses `PARTICLE_NO_PREPROCESSOR`


kennethlimcp:
@mumblepins can you point me to where this change/update was written?

I see that the current builder is still - https://github.com/spark/buildpack-wiring-preprocessor/blob/13dda1ea0d46dc82b63ef0d7934a51ab0c9a6b8f/cparser/processor.js#L43

mumblepins:
Huh.  That's interesting (always the prelude to something breaking, right?  :smile: ).  I was basing this purely off of the Web IDE results, and screwed up.  Sorry!  On a different note, the road to Spark calls to Particle calls has some potholes. Sigh.

PR #319, opened 2016-03-17
Add system-button and reset-button template helper

monkbroc:
- `{{system-button}}` inserts SETUP or MODE depending on the device
- `{{reset-button}}` inserts RESET or RST depending on the device
- Replace SETUP/MODE/RESET/RST in the docs with new template helpers

This is a more complete version of #311 so it closes #311


monkbroc:
Merged as c5c54f87

PR #322, opened 2016-03-21
Clarify the return value of WiFi.scan()

monkbroc:
Per spark/firmware#917


jamesr66a:
I think "added" should be changed to "written", since adding to some collection implies the existing contents are preserved-which I don't believe is the case here

monkbroc:
Good point

PR #323, opened 2016-03-21
Update EEPROM documentation for atomic operations

monkbroc:
Documentation for spark/firmware#871
- [x] Proofread

### DIFF
--- a/src/content/reference/firmware.md
+++ b/src/content/reference/firmware.md
@@ -5605,38 +5605,129 @@ from the cloud, and setting the seed is left to up you.
 
 ## EEPROM
 
-The EEPROM emulator allocates a region of the device's built-in flash memory to act as EEPROM.
+EEPROM emulation allocates a region of the device's built-in Flash memory to act as EEPROM.
 Unlike "true" EEPROM, flash doesn't suffer from write "wear" with each write to
 each individual address. Instead, the page suffers wear when it is filled. Each write
 will add more data to the page until it is full, causing a page erase.
 
-The EEPROM functions can be used to store small amounts of data in flash that
+The EEPROM functions can be used to store small amounts of data in Flash that
 will persist even after the device resets after a deep sleep or is powered off.
 
+On the Photon/Electron, Flash pages are rated for 10,000 erases and it takes 20,000 byte writes to
+fill the EEPROM pages, so expect to be able to write 200,000,000 bytes to the EEPROM in the lifetime
+of the Photon/Electron.
+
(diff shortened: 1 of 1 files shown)


src/content/reference/firmware.md
```diff
@@ -5666,70 +5759,57 @@ uint8_t val = 0x45;
 EEPROM.write(addr, val);
 ```
 
-### update()
-This method is similar to `EEPROM.write()` however this method will only write data if the cell contents pointed to by `address` is different to `value`. This method can help prevent unnecessary wear on the EEPROM cells.
+When writing more than 1 byte, prefer `put()` over multiple `write()` since it's fater and it ensures
```
technobly: fater -> faster

src/content/reference/firmware.md
```diff
@@ -5666,70 +5759,57 @@ uint8_t val = 0x45;
...
+Erasing a page of Flash pauses processor execution (including code running in interrupts) for 500ms since
+no instructions can be fetched from Flash while the Flash controller is busy erasing the EEPROM
+page.
 
-`address` is the address (int) of the EERPOM location to write to
-`object` is the object data to write
+`hasPendingErase()` lets the application developer check if a full EEPROM page needs to be erased.
+When the application determines it is safe to pause processor execution to erase EEPROM it calls
+`performPendingErase()`. You can call this at boot, or when your device is idle if you expect it to
+run without rebooting for a long time.
```
technobly, line 5790: Does this work with multi-threading?  How long should a full page erase take?  Is there any reason we don't handle this automatically with the user option to handle manually?

monkbroc, line 5790: There are 2 pages used for EEPROM emulation. One takes 300ms to erase and one takes 500ms to erase. It doesn't matter if you have multi-threading on or off since the processor just stops executing instructions while erasing a page of flash. There's nothing you can do about this on the STM32F205. You can see this for example when you DFU flash. The yellow blinking pauses a few times while the Flash is being erased.

The page erase is handled automatically with the user option to handle it manually. If it does not come across clearly I can wordsmith this section a bit.

technobly, line 5790: Ok cool, that makes sense but I didn't pick up on that while skimming the added sections for some reason.  Just a note that those functions are optional, and how they become mandatory would be good.  Might be worth adding the known times in there for erase, and mentioning that the erase process is blocking.

monkbroc:
Merged as  e1c7c4e2

Issue #304, opened 2016-03-02
number of var/func is out dated


kennethlimcp:
It's stated here:

```
Currently the application supports the creation of up to 4 different cloud functions.
```

https://docs.particle.io/reference/firmware/electron/#particle-function-

But the thread says:

```
the 0.4.x firmware supports an unbounded number of functions and variables. 
```

http://community.particle.io/t/can-you-change-how-many-web-functions-variables-you-can-register/20635/6

hkmoca:
I've made the change of this issue in this pull request:

https://github.com/spark/docs/pull/324

PR #324, opened 2016-03-22
number of var/func is out dated #304

hkmoca:
fixed issue #304 changed description of the current supported var/func on Particle.

### DIFF
--- a/src/content/reference/firmware.md
+++ b/src/content/reference/firmware.md
@@ -113,7 +113,7 @@ bool success = Particle.function("funcKey", funcName);
 //                     (max of 12 characters long)
 ```
 
-Currently the application supports the creation of up to 4 different cloud functions.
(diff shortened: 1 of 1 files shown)


monkbroc:
Thanks. I was updating this section of the docs today. I'll include your change.

monkbroc:
Before writing in the docs that there is no limit, I'd like to test what happens when you register more variables than actually fits in the description message.

hkmoca:
hi man! how are you doing? 
Is there anything that i could help you to merge this one?

monkbroc:
Since the number of functions and variables is limited by the size of the discovery message I'd rather not say "unlimited" in the docs, otherwise somebody might create a Particle.variable for every variable in their software. I need to check the practical limit.

monkbroc:
I've verified that up to 20 variables and 15 functions fit in the "describe" message sent by the device to the cloud. I'm updating the docs to reflect that in #397.

PR #337, opened 2016-03-27
Update shields link

kennethlimcp:
Use the `particle-shields` link as per the title of the document instead of the old `photon-shields`

Would appreciate a review and merge. @technobly @monkbroc


monkbroc:
Thanks Kenneth. Looks good. I'm merging.

PR #340, opened 2016-03-28
Corrects misnamed field in event stream data object

lourd:
It's confusing because the function parameter name in many places is `deviceId`, but real data from the event stream shows that it's `coreid` that comes through.


lourd:
Also the same correction I made a couple months ago in https://github.com/spark/docs/pull/281

monkbroc:
This is confusing for sure :confounded:

Thanks for the correction.

PR #341, opened 2016-03-28
Example improvement

davidgatti:
This example wasn't copy&paste compatible, and the `Udp.begin(8888);` part was missing. Without this, no data will be sent. It is true that this line was mentioned in a example before, but not here. This caused confusion for me, and other people on the internets. 

I hope this helps.

### DIFF
--- a/src/content/reference/firmware.md
+++ b/src/content/reference/firmware.md
@@ -4271,6 +4271,10 @@ Sends a packet, unbuffered, to a remote UDP peer.
 
 // SYNTAX
 
+UDP.sendPacket(buffer, bufferSize, remoteIP, remotePort)
+
+// Example Usage
+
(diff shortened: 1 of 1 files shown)


src/content/reference/firmware.md
```diff
@@ -4271,22 +4271,24 @@ Sends a packet, unbuffered, to a remote UDP peer.
...
-int remotePort = 22;
+IPAddress remoteIP(192, 168, 1, 100);
+int port = 1337;
 
-// fill the buffer with goodness
-// ...
+void setup() {
 
-// now send the buffer as a packet
+    // Requierd for two way communication
```
kennethlimcp: Required

monkbroc:
Thanks for your contribution! I've merged it.

PR #347, opened 2016-04-05
Made the safe mode more clear

davidgatti:
I think this new feature requires its own section :) Right now it is hard to spot. I hope this helps.

### DIFF
--- a/src/content/guide/getting-started/modes.md
+++ b/src/content/guide/getting-started/modes.md
@@ -58,7 +58,9 @@ If your {{device}} is blinking magenta, it is currently loading an app or updati
 
 {{{device-animation device "blink" "magenta" }}}
 
-If your latest code is crashing your Particle, this mode is for you! Hold down the {{#if photon}}`SETUP`{{/if}}{{#unless photon}}`MODE`{{/unless}}, then click the reset button once. When your Particle starts blinking in magenta, imediatly release the {{#if photon}}`SETUP`{{/if}}{{#unless photon}}`MODE`{{/unless}} button. If you do this step correctly, the device will enter in safe mode, and won't execute your old code.
+You should put your Particle in safe mode when you want to perform a safe OTA (no code is running on the device), or when your latest code is crashing the device.
+
+Hold down the {{#if photon}}`SETUP`{{/if}}{{#unless photon}}`MODE`{{/unless}}, then click the reset button once. When your Particle starts blinking in magenta, immediately release the {{#if photon}}`SETUP`{{/if}}{{#unless photon}}`MODE`{{/unless}} button. If you do this step correctly, the device will enter in safe mode, and won't run your code.
 
 ### Looking For Internet
 
--- a/src/content/guide/getting-started/modes.md
+++ b/src/content/guide/getting-started/modes.md
@@ -48,15 +48,17 @@ Here's the typical pattern of a {{device}} after power up.
 
 When it is breathing cyan, your {{device}} is happily connected to the Internet. When it is in this mode, you can call functions and flash code.
 
-
 ### OTA Firmware Update
 
 {{{device-animation device "blink" "magenta" }}}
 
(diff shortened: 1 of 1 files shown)


src/content/guide/getting-started/modes.md
```diff
@@ -48,15 +48,17 @@ Here's the typical pattern of a {{device}} after power up.
...
 {{{device-animation device "blink" "magenta" }}}
 
 If your {{device}} is blinking magenta, it is currently loading an app or updating its firmware. This state is triggered by a firmware update or by flashing code from Particle Dev or Particle Build. You will often see this mode when you connect your {{device}} to the cloud for the first time.
 
+### Safe mode
+
+{{{device-animation device "blink" "magenta" }}}
 
-Note that, if you enter this mode by holding {{#if photon}}`SETUP`{{/if}}{{#unless photon}}`MODE`{{/unless}} on boot, blinking magenta indicates that letting go of the {{#if photon}}`SETUP`{{/if}}{{#unless photon}}`MODE`{{/unless}} button will enter safe mode to connect to the cloud and not run application firmware.
+If your latest code is crashing your Particle, this mode is for you! Hold down the {{#if photon}}`SETUP`{{/if}}{{#unless photon}}`MODE`{{/unless}}, then click the reset button once. When your Particle starts blinking in magenta, imediatly release the {{#if photon}}`SETUP`{{/if}}{{#unless photon}}`MODE`{{/unless}} button. If you do this step correctly, the device will enter in safe mode, and won't execute your old code.
```
kennethlimcp: Typos here and also, safe mode is not just for devices with crashing code. This ends up making it less clear.

davidgatti: Mmm... so how can we make it more clear? Because for me how it is explained now, it is unclear, the word `safe mode` is hidden within the sentence, and I, for example always miss it while skimming through the document. 

Also, you say this mode is not only for crashing code, can you tell me what other use cases there are?

kennethlimcp: The main goal of safe mode is to a device in the state where the user code is not running at all. This can be deliberate through code or manually through the physical buttons.

We can always add a note to say that this is useful when your device firmware is crashing and not maintaining :cloud:  connection. Placing the device in safe mode allows you to OTA new firmware without any user code running.

I would really recommending proof-reading (editors should have dictionary feature enabled as well) PRs too.

davidgatti: I'll read once again what I wrote. 

In the mean time. I'm still bit confused. Why would I want to set the Particle in safe mode if everything is ok with my code?

kennethlimcp: There might be cases that you want to place the device in safe mode so that none of your firmware/project is running.

For example, you are controlling a water heater and you perform an OTA firmware update halfway. The code hangs, your Photon is 1000miles away and OTA firmware didn't work. The water heater keeps firing and eventually lead to a mishap.

Deliberately handling the peripherals status properly before entering Safe Mode ensures that no user code is running during an OTA. If the OTA fails, the device will simply reset and run the old firmware. Hope this helps.

davidgatti: In this case could we write the following:

You should put your Particle in safe mode when you want to perform a safe OTA (no code is running on the device), or when your latest code is crashing the device.

What do you think?

davidgatti:
I think now this is nice :) what do you think @kennethlimcp

moors7:
I could be mistaken, but doesn't Safe Mode [already have a section of its own](https://docs.particle.io/guide/getting-started/modes/photon/#safe-mode)? Stating it at the OTA firmware upgrade part is a simply pointer on what happens if you hold Mode on setup, and how you can recognize whether or not you're in Safe Mode. Personally I'd just make the mention of "Safe Mode" over there a link to the already existing docs.

davidgatti:
😮 how did I miss that. I'm so sorry for wasting your time guys.

PR #358, opened 2016-04-14
fixes the redirect_uri to be https. otherwise we get an error

rayjanwilson:
original docs said the `redirect_uri=http://www.pricle.io/setup`
and curl would return:

```
{
  "ok": false,
  "errors": [
    "The \"redirect_uri\" parameter must be a valid HTTPS URL"
  ]
}
```

adjusting the redirect_uri to https results in the expected behavior


kennethlimcp:
@rayjanwilson closing this since it has been superseded by this --> https://github.com/spark/docs/pull/343/files

Thanks!

PR #361, opened 2016-04-16
Confirming That The Webhook Successed

gdeflaux:
Hi. I thought would be helpful to add this to the documentation as I didn't find it anywhere and needed it. It's yours to update/correct and publish if valid. Cheers


monkbroc:
Thanks. I included your suggestion.

PR #363, opened 2016-04-18
fix EEPROM default value example+doc mismatch

tedder:
The documentation says, in four places, the default value is 0xFF. The example uses 0xFFFF. The former makes sense, given 8 bits in a byte. That means `0xFFFF` would fail (unless it happens to work by ignoring the upper bits).

### DIFF
--- a/src/content/reference/firmware.md
+++ b/src/content/reference/firmware.md
@@ -6161,7 +6161,7 @@ from the type of object.
 int addr = 10;
 uint16_t value;
 EEPROM.get(addr, value);
-if(value == 0xFFFF) {
+if(value == 0xFF) {
   // EEPROM was empty -> initialize value
   value = 25;
 }


tedder:
Travis failure is `npm`, not my change.

monkbroc:
Since the example code uses `uint16_t value` then 2 bytes of EEPROM will be read into `value` giving `0xFFFF` when the EEPROM is empty.

You can try to run the example code with empty EEPROM (you can use `EEPROM.clear()` first) and you'll see that `value` is `0xFFFF`.

tedder:
Aha- I didn't catch that you were taking something bigger than a byte out. What do you think about changing it to a byte, since all the documentation is at the byte level?

monkbroc:
I'd rather show that you can get any type of value to discourage people from reading individual bytes then reassembling them into larger values. Doing that takes more time and is not robust in case of power loss during writes.

tedder:
That makes sense, given the warnings about reading tiny blocks. What about two examples, then? Or at least a comment explaining the length of the uint16?

monkbroc:
Sure. Just did it in https://github.com/spark/docs/commit/87b0316ec311dd49fa0232c544f0d267a78058cf

PR #368, opened 2016-04-19
ApplicationWatchdog docs

m-mcgowan:


monkbroc:
I'll add our usual `// SYNTAX` and `// EXAMPLE USAGE` on the right hand side

technobly:
- Will the watchdog callback be called if code is off running in a hard loop with critical priority somewhere?
- What is the range for: "A default stack size of 512 is used for the thread. The stack can be made larger or smaller as needed."
- May be unclear why a user would need to change the stack size, maybe include example with the explanation of that: `ApplicationWatchdog wd(60000, System.reset, 1024);`

monkbroc:
Also, I guess it's not necessary for system threading to be enabled for this to work, right?

m-mcgowan:
The only time the watchdog isn't called is if the user code blocks in an interrupt which pre-empts systick, preventing task switching.  Critical priority threads are still time-sliced so the watchdog would still be serviced. 

The stacksize can be whatever - 512 worked for System.reset, and smaller values might work, but I didn't spend time investigating. Larger values work too of course, but this is just a waste of memory allocating more stack than is needed. Changing the stack size is only needed if the timeout function needs a larger stack.

SYSTEM_THREAD() is not needed - FreeRTOS is always active and time-slicing.

technobly:
Thanks for the explanation!

Since these compile without warning or error, is there a good way to know when more memory is needed for the callback function?

```
ApplicationWatchdog wd(10000, System.reset, -1);
ApplicationWatchdog wd(10000, System.reset, 0);
ApplicationWatchdog wd(10000, System.reset, 1);
```

monkbroc:
I'd say to leave the stack size stuff unsaid. The default will be used in most cases.

src/content/reference/firmware.md
```diff
@@ -5759,6 +5759,36 @@ timer.dispose(); // stop and delete timer from timer list.
...
+- A default stack size of 512 is used for the thread. The stack can be made larger or smaller as needed.
+
+Example:
+
+```
+ApplicationWatchdog wd(60000, System.reset);
+// reset the system after 60 seconds if the application is unresponsive
+
+```
+The application watchdog requires interrupts to be active in order to function.  Enabling the hardware watchdog in combination with this is recommended, so that the system resets in the event that interrupts are not firing.
```
monkbroc, line 5785: Is the hardware watchdog documented anywhere? It is enabled by default, right?

technobly:
I mainly ask this for product creators... when setting this up if it doesn't have enough memory allocated maybe it just ends in an SOS fault when AWDT occurs?  These limits are something that should be tested before launching a product.  In theory this callback is never called, so it may be easy to write code that assumes it will be called properly if that day ever comes.  However, someone may never attempt to test that it will save the day :)  These are the kinds of things we should put in the docs for product robustification.

PR #372, opened 2016-04-23
fix minor grammatical error

tedder:
powered cycled -> power cycled


tedder:
Unsure why it's failing Travis with a localhost URL. The URL is obviously valid.

kennethlimcp:
@tedder thanks :)

PR #388, opened 2016-04-29
Stm32fxx url fix

tedder:
update URL that is 404ing, seen in #387


kennethlimcp:
fixed here: https://github.com/spark/docs/commit/0ae9f096dabbcc2323a588f7c00786f54b5e8b66

PR #389, opened 2016-04-29
Callback function signature

tedder:
Shows two details that are easy for newer programmers to miss- the importance of the return type and arguments.


monkbroc:
Thanks! I edited your changes a bit and merged them.

PR #398, opened 2016-05-07
edit link for firmware

m-mcgowan:
adds a handlebar helper `edit-link` to produce the href for editing a file. This is used to direct `firmware.md` to the correct github file (since firmware.md is a symlink.). I chose the quick and simple approach (hard-coding), rather than resolving symlinks, which can be dangerous.

### DIFF
--- a/templates/helpers/edit-link.js
+++ b/templates/helpers/edit-link.js
@@ -1,0 +1,14 @@
+module.exports = function(branch, path, file) {
+	/* todo - we could check if the file is a symlink and rewrite the path to the target. 
+	  For now we take the simple, trustworthy appraoch. */
+	var content;
+	var base;
+	if (file==="firmware") {
+		content =  "firmware/firmware.md";
+	}
+	else {
+		content = "src/content"+path+file+".md";
+	}
+	base = "https://github.com/spark/docs/tree/"+branch+"/";
+	return base + content;
+}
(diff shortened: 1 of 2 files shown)


templates/helpers/edit-link.js
```diff
@@ -0,0 +1,12 @@
+module.exports = function(branch, path, file) {
+	/* todo - we could check if the file is a symlink and rewrite the path to the target. 
+	  For now we take the simple, trustwirthy appraoch. */
```
monkbroc: trustwirthy: typo

monkbroc:
The style in this repository is to use semicolons at the end of Javascript statements. Can you add semicolons to `edit-link.js`?

templates/helpers/edit-link.js
```diff
@@ -0,0 +1,12 @@
...
+	if (file=="firmware") {
```
monkbroc: Use `===` instead of `==` in Javascript

templates/helpers/edit-link.js
```diff
@@ -0,0 +1,12 @@
+module.exports = function(branch, path, file) {
+	/* todo - we could check if the file is a symlink and rewrite the path to the target.
```
monkbroc, line 2: Declare `var content, base;` at the top of the function otherwise those variables will become global variables (yay, Javascript!)

m-mcgowan:
Thanks for the suggestions - I've made the changes requested.  How about we add jslint to the build?

monkbroc:
> How about we add jslint to the build?

We have eslint in our newere repos. We can add that as we update this repo to ES6 syntax.

PR #407, opened 2016-05-16
Update 'connect with your smartphone' section

AlexSneedMiller:
Making it a little bit more detailed to include pressing on the plus icon for starting the process to start connecting the device. 

Perhaps this is obvious to Android users as part of Google's Material Design.


monkbroc:
Yeah, that wasn't clear. Thanks for the suggestion.

PR #414, opened 2016-05-31
One carriage return isn't enough to make a new paragraph

embedded-creations:


monkbroc:
Just can't get enough of those carriage returns!

PR #416, opened 2016-06-06
Added instructions to get out of hard fault

nrobinson2000:
Making change due to ctmorrison's request
https://community.particle.io/t/hard-fault-recovery/23505/8?u=nrobinson2000

### DIFF
--- a/src/content/guide/getting-started/modes.md
+++ b/src/content/guide/getting-started/modes.md
@@ -375,7 +375,6 @@ If your {{device}} is stuck in Hard fault, try putting your {{device}} into DFU 
 particle flash --usb tinker
 ```
 
-
 **Out of heap memory (8 blinks between 2 SOS patterns)**
 
 {{{device-animation device "sos" 8 }}}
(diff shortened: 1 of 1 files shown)


monkbroc:
I'm not sure we want to add this. We prefer telling people to put their device in safe mode instead of DFU because then they can reflash a new app from the cloud in case their current app crashes with an SOS.

nrobinson2000:
Maybe we should then give the instruction to put the device into Safe mode in that part of the page then.

kennethlimcp (approved):

technobly:
I'd be happy with these changes :)

<pre>
If your {{device}} is stuck in Hard Fault or other SOS code, first try recovering with [Safe Mode](/guide/getting-started/modes/electron/#safe-mode) and flashing Tinker OTA with the CLI to see if it was something recently added in your user application.

```
particle flash <device_name> tinker
```

If it's not possible to enter Safe Mode, your system firmware may be corrupted.  Try putting your {{device}} into [DFU Mode](/guide/getting-started/modes/photon/#dfu-mode-device-firmware-upgrade-) and flashing the system firmware locally over USB (and optionally flash the Tinker application) with:
 
```
particle update

particle flash --usb tinker
```

</pre>

PR #417, opened 2016-06-06
[community] Add a page about po-util to "Tools and Features"

nrobinson2000:
I was thinking that we could add a page about po-util, the script that I have been developing into the Particle documentation.  I have been using it for all of my Particle development and I think that other users will benefit from having it in the documentation.

### DIFF
--- a/src/content/guide/tools-and-features/po-util.md
+++ b/src/content/guide/tools-and-features/po-util.md
@@ -32,7 +32,7 @@ order: 8
 You can download **po-util** from [GitHub](https://github.com/nrobinson2000/po-util), or install the latest version with:
 
 ```
-curl po-util.com/download
+curl po-util.com/download | bash
 ```
 
 ## Examples
@@ -99,7 +99,7 @@ po electron ota YOUR_DEVICE_NAME
 ```
 {{/if}}
 
-# Support
+## Support
 
 Nathan Robinson can be contacted on the [Particle Forums](https://community.particle.io/users/nrobinson2000).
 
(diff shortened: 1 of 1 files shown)


monkbroc:
I agree that installing all these tools is hard. One of my current tasks is creating installers for the Particle tools. I'm starting with the CLI and later I'll do one for the local compiler. I'll take a look at your Linux and OSX script, but I don't think I want to include all the functionality into a tool installer. Some of that functionality like local build belongs in the CLI in the long run.

nrobinson2000:
> I don't think I want to include all the functionality into a tool installer.

Why not?  It works smoothly for me.  Its also very easy to install.

> Some of that functionality like local build belongs in the CLI in the long run.

True.  But I think that for now my script is good enough.

nrobinson2000:
Po-util is more of a wrapper than a tool.  It makes it easier to do difficult tasks like installing the necessary build tools and building locally.

kennethlimcp:
Happy to add this under `community.md` as a link but not a page itself

nrobinson2000:
Yes, a link would be perfect.

nrobinson2000:
@kennethlimcp 
Any update on adding a link to po-util in the docs?  Do I have to create a pull request for this?

technobly:
@nrobinson2000 I tagged this issue with the **[community]** tag so that when we have a page figured out for Community projects, libraries, tools we will resolve these related issues at that time.  @cmsunu28 may be working on this page already (or in the future) if I remember correctly.

nrobinson2000:
That sounds great.  I did just make this pull request however.  https://github.com/spark/docs/pull/538

monkbroc:
I merged in #538 so I'll close this one.

PR #430, opened 2016-06-27
Update firmware.md

rcaudill:
Clarify location of Serial2 pins on non core devices


harrisonhjones:
This appears to only apply to the Photon. What about the Electron?

rcaudill:
You are correct, this only applies to Photon.

monkbroc:
Thanks. I resolved the conflict and merged the changes.

PR #437, opened 2016-07-07
Additional details on passing multiple values to a webhook

sbirdranch:
I added additional verbiage and a snippet illustration on how to pass multiple values to a web hook, and then how to address them as dynamic values.


monkbroc:
I added a new reference for webhooks with this information. Thanks!

https://github.com/spark/docs/commit/84a76d516607b9855ab6062f0988c17d35885c68

PR #438, opened 2016-07-11
Update modes.md

taylorsloan:
I'm on Win 7 and couldn't flash using dfu-util even though I had all the drivers installed. Even though my Electron is visible as a dfu device in my device manager, dfu-util will not detect it as a dfu device if it is connected via a usb3 port. I connected my electron via an old usb hub and it worked. Telling people about this prior will save them time.


monkbroc:
Hi @bropane, sorry you had trouble getting DFU util to connect to your Electron. dfu-util seems to work ok on most USB 3 ports, including on my Windows 7 64-bit computer. Could you please reword to say something like "try a USB 2 port if a USB 3 port doesn't work"?

PR #442, opened 2016-07-18
Fix typo in power shield documentation

AlexSneedMiller:
Fixed plural issue and made 'led' into 'LED' for consistency with the rest of the document


kennethlimcp:
@AlexSneedMiller thanks!

PR #451, opened 2016-08-08
wrong url

dirkk0:
You need to access github with 'git clone https://github.com/spark/particle-cli.git' otherwise you'll get an error.


kennethlimcp:
@dirkk0 thanks! I will fix it since the `:` needs to be replaced to `/` as well

PR #454, opened 2016-08-09
blinking orange fix

mohitbhoite:
Suggest a fix when the device exhibits blinking orange state - indicating bad server keys.


technobly:
I have some ideas about this as well... We need to fix up the LED signaling around all of these errors to make the specific issue easier to diagnose. While the server key is typically a problem, I've seen a red blip more associated with it. Not to say orange blinking is not possible as well.

PR #467, opened 2016-09-17
Use `hostname` instead of `URL` terminology

dougalcampbell:
It's pedantic, but `URL` really isn't correct in this context (`TCPClient::connect()`). A URL can consist of several pieces, including a scheme, hostname, port, path, etc. If you tried to provide a valid URL like `http://www.google.com/` to `TCPClient::connect()`, it would be an error. We're really just using a host name here.

### DIFF
--- a/firmware/firmware.md
+++ b/firmware/firmware.md
@@ -4587,13 +4587,13 @@ Connects to a specified IP address and port. The return value indicates success 
 // SYNTAX
 client.connect();
 client.connect(ip, port);
-client.connect(URL, port);
+client.connect(hostname, port);
 ```
 
 Parameters:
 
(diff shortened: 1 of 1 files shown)


kennethlimcp:
@technobly ?

rickkas7:
The change is correct. TCPClient::connect() takes neither a URL nor a domain name; the parameter is a hostname.

kennethlimcp:
i guess we follow Arduino convention but always open to improvements:

From https://www.arduino.cc/en/Reference/ClientConnect:

```
URL: the domain name the client will connect to (string, ex.:"arduino.cc")

```

technobly:
Thanks all!

dougalcampbell:
Thanks folks. It's just one of those things I tend to be pedantic about. :)

Issue #474, opened 2016-09-29
firmware-upgrades.md doesn't work for Electrons


harrisonhjones:
docs/src/content/support/troubleshooting/firmware-upgrades.md doesn't really work for Electrons. Need to copy over and update the relevant Photon instructions.

alan707:
this issue should be closed with PR #477

PR #477, opened 2016-10-05
added firmware update case for Electron

alan707:
fixes issue #474


monkbroc:
Thanks for your contribution!

alan707:
No problem, @monkbroc happy to keep the wheels turning!

PR #478, opened 2016-10-06
Add B0-B5, C0-C5 to GPIO list in note for electron

mfogel:
There's an offhand list of GPIO's in two notes in the firmware docs for the Input/Output functions. These lists of GPIO's are currently photon-specific. This PR adds the additional electron GPIO's to the lists, if the `electron` flag is set.


monkbroc:
Thanks for add this!

PR #479, opened 2016-10-11
New section Firmware Releases

technobly:
Adds these sections to Firmware Reference in Docs:
- Firmware Releases
  - Firmware Release Process
  - Github Release Notes
  - Programming and Debugging Notes

Github Release pages will now link to the Programming and Debugging Notes section with a dynamic URL such as https://docs.particle.io/reference/firmware/photon/?fw_ver=0.5.3&cli_ver=1.17.0&electron_parts=2#programming-and-debugging-notes instead of statically created for each firmware release.


templates/partials/scripts.hbs
```diff
@@ -28,9 +28,9 @@
 <script src='/assets/js/docs.js'></script>
 <script src='/assets/js/ga.js'></script>
 <script src='/assets/js/device-animation.js'></script>
+<script src='/assets/js/firmwareversions.js'></script>
```
jme783: Would it be possible to load this script on only the pages you need instead of on every one? Sometimes weird JS things happen when some jQuery expects content that isn't there

jme783 (approved):
Overall, looks like this will do the trick! Using jQuery in this way makes me a tad bit nervous from a maintainability and rigidity standpoint, but this seems to be a short-term solution to unblock the firmware team.

In the long term, I am very excited about coming up with a flexible and straightforward way to add versioning to our documentation.

monkbroc:
Let's not hold up the release but I did mention to @stefan-moraru that it should be possible to use a Handlebar block helper like this instead of using markers that will be removed by jQuery. We just haven't gotten around to pairing on this...

```
{{#section "release-notes-0.5.3}}
# Notes
More markdown
{{/section}}
```

templates/partials/scripts.hbs
technobly: Stefan and I got that worked out @jme783 so now the script will only load on reference pages and only run on the firmware.md page.

stefan-moraru (approved):
Looks good

technobly:
Most of the latest firmware releases and prereleases now link back to this new Docs feature: https://github.com/spark/firmware/releases

Issue #19, opened 2015-05-04
Open Datasheet Errata


technobly:
> Datasheet errata will be added to this issue when found.  Datasheets will be updated with the listed version after items are checked off.  Some errata may persist between revisions, and they will be moved up to the next proposed revision.
# E Series Datasheet (v002) - Not Released Yet
- [ ] Mention the use of optional SPI FLASH and the associated GPIO dependencies.
- [ ] Clarify on the use of the GND pads on the bottom
- [ ] Mention operation in below 0C temperatures for LiPo batteries

# Photon Datasheet (v016)
- [x] Updated External Coexistence Interface section (currently unsupported)

# P1 Datasheet (v010)
- [x] Change "Paid tiers of Particle Cloud start when you create a product with more than 25 devices" to say "Particle provides free access to the Particle Cloud for prototyping and paid tiers for production".
- [x] Updated External Coexistence Interface section (currently unsupported)

# P1 Datasheet (v009)
- [x] Added part number for 1MB external SPI flash (MX25L8006E)

# P1 Datasheet (v008)
- [x] EEPROM emulation bank 2 (64 KB) - remove only 16k used
- [x] Replace STM32 pin names with Particle/Wiring API names to keep everything intuitive and high level.  Leave STM32 pin numbers for deep dives. #545
- [x] Make Image of P1 Block Diagram Snazzy
- [x] CAN bus pins should be noted as 3.3V tolerant (technically they are 5V tolerant on the RX but since you wouldn't operate them with a 5V transceiver it's proper to classify them as 3.3V)
- [x] Fix and clarify pin descriptions https://github.com/spark/docs/pull/399 (make sure to review P1 and Photon datasheets as well)
- [x] Add note to clarify why MICRO_SPI1_SS is not hooked up to DAC/A6 (PA4) https://github.com/spark/docs/issues/469
- [x] update DCT Layout https://docs.particle.io/datasheets/p1-datasheet/#dct-layout notably adding radio_flags, and any other new sections.
- [x] JTAG pin P1 module number is wrong.  "D4	JTAG_TDO PB3 54 MICRO_JTAG_TDO Floating" should be "D4 JTAG_TDO PB3 52 MICRO_JTAG_TDO Floating"  (54 -> 52)
- [x] Add warning about powering the Photon from long wires on USB and VIN, and give advice on what components can be added externally to mitigate this.  From Richtek datasheet:

[...]

PR #482, opened 2016-10-27
Minor typo fix, though one that drove me crazy

eely22:


monkbroc:
Thanks for bringing this up.

@technobly Could you confirm that this is the correct part number for the Electron and if so please add it to the datasheet errata? https://github.com/spark/docs/issues/19

technobly:
That is correct, I'll get this fixed up.  Thanks @eely22 !

technobly:
https://github.com/spark/docs/pull/483 merged, give it a few minutes and it will be live.

PR #490, opened 2016-11-03
Fix required frequency of calling Particle.process() for electrons

mfogel:
Default keep alive for electrons:
https://github.com/spark/firmware/blob/v0.5.3/communication/src/dtls_protocol.cpp#L14

Default keep alive for others:
https://github.com/spark/firmware/blob/v0.5.3/communication/src/lightssl_protocol.h#L64


monkbroc:
I think we still want to tell people not to block for extended periods of time otherwise the device will appear unresponsive from the cloud for Particle.function, Particle.variable and flashing new code.

monkbroc:
I'm going to close this one to avoid giving the impression that it's OK to block the main loop for many minutes without calling `delay()` or `Particle.process()`

PR #492, opened 2016-11-08
Broken link for cellular provider

alexbjorlig:
The real link is https://www.particle.io/pricing#cellular-data


monkbroc:
Thanks!

PR #500, opened 2016-11-18
corrected a statement

alan707:
This statement is not correct:
>You can add these two "macros" anywhere in your application code. 

If you add PRODUCT_ID and PRODUCT_VERSION above the #includes it will not compile since it is missing the 'Application.h'


monkbroc:
I tweaked the wording a bit. Thanks for the clarification.

PR #513, opened 2016-12-02
Feature/faq rick

rickkas7:
Contains FAQ entries for:

Connectivity
  Finding your device ID
  3rd-party SIM cards

Tools
  Installing dfu-util
  JTAG and SWD
  Installing curl (for Windows)

Will has looked over the changes already.


src/content/faq/tools/installing-dfu-util.md
```diff
@@ -0,0 +1,353 @@
...
+sudo yum install dfu-util
+```
+
+Arch:
+
+```
+sudo pacman -Sy dfu-util
+```
+
+You will also want to add a udev rule. This allows dfu-util to work without requiring sudo, which is convenient, and also necessary when using certain CLI and local build commands.
```
technobly: This is done automatically now with the CLI if you run `particle update` or `particle flash --usb ...`, but would probably be good to have a manual way of adding it as well.

src/content/faq/tools/jtag.md
```diff
@@ -0,0 +1,369 @@
...
+
+Select **Connect** from the **Target** menu or use the toolbar icon. It should look something like this:
+
+![ST-LINK Connect](/assets/images/jtag-02connect.png)
+
+The bootloader is different for each type of device and changes infrequently. Download the appropriate one for your device:
+
+- [Photon](https://github.com/rickkas7/particle_jtag/blob/master/files/bootloader-photon.bin?raw=true) (0.4.9)
+- [P1](https://github.com/rickkas7/particle_jtag/blob/master/files/bootloader-p1.bin?raw=true) (0.4.9)
+- [Electron](https://github.com/rickkas7/particle_jtag/blob/master/files/bootloader-electron.bin?raw=true) (0.4.8)
```
technobly: I'd like to make this more future-proof.  Those Photon/P1 bootloaders kind of feel like they might already be outdated if they are not v7.

technobly:
Great info Rick!

src/content/faq/tools/installing-dfu-util.md
rickkas7: That explains it! I didn’t need one the first time I tested it, but when I redid it on a clean machine to verify the text and take the screenshots, I ran into needing a udev rule because I hadn’t installed the CLI. I can fix the text to explain that.

PR #519, opened 2016-12-13
Rework the example to be more relevant for the Pi.

monkbroc:
Tweaked the blink LED, added the CPU temperature script execution and removed non-relevant examples.


monkbroc:
This is the Fritizing part I used:
https://github.com/spark/hardware-libraries/blob/feature/raspberry-pi-particle/Fritzing/parts/Raspberry%20Pi%203%20(Particle).fzpz

PR #533, opened 2016-12-29
[community] Added LogHandlers section

harrisonhjones:


firmware/firmware.md
```diff
@@ -9802,7 +9802,8 @@ void loop() {
 
 The library provides the following log handlers:
 
-`SerialLogHandler`
+- `SerialLogHandler`
+- Additional community-supported log handlers can be found further below.
```
technobly, line 9806: Adding this comment here feels out of place, as a way to direct people elsewhere away from the official logging docs.  I don't think we should start adding non-official libraries in between official docs.  I see the benefit of this papertrail logger, but I think we need a separate Community section of the docs.  If it's unofficial it means we don't maintain it, and run the risk of these items being confusing to new users if they are not maintained well.  There are hundreds of libraries, and you can imagine how confusing it will be if we started adding them everywhere in the firmware reference.

technobly (changes requested):

firmware/firmware.md
harrisonhjones: So, it sounds like you don't have an issue with such a comment there but you don't like the actual list of community loghandlers further below? I'd be happy to create a community section of the docs and then link to it from here. Thoughts? Not sure how easy a new section will be to create but I can look into it.

technobly: I'm not sure we want to be linking from official docs to community creations either, but a place for community libraries/projects/example applications would be a nice addition to the Docs.  I know we've talked about adding this soon, so maybe we should hold off on adding this for now to see what the format of this new section will look like.  I'll tag this with a new milestone of "community" so we can keep track of the PRs/issues that represent community additions.

PR #534, opened 2016-12-30
Updated sample Swift code to Swift 3 and to reflect current state of the API.

stardust66:
Changes:
-println() to print()
-NSError to Error
-if let statements turned into if variable != nil
-loginWithUser(...) changed to login(withUser: ...)
-publishEventWithName(...) changed to publishEvent(withName: ...)

The old version leads to errors or warnings when used in the newest version of Xcode. These changes fix them.


monkbroc:
Thanks!

PR #536, opened 2016-12-31
Auth section qoutes error

Jjbasso:
Seemed like the token needed to be in single quotes.  Not sure for the compile portion of the code.

### DIFF
--- a/src/content/reference/javascript.md
+++ b/src/content/reference/javascript.md
@@ -97,7 +97,7 @@ devicesPr.then(
 Call a function in device
 
 ```javascript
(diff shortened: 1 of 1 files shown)


technobly:
Did you implement this part?
`var token; // from result of particle.login`

monkbroc:
I clarified where to get the `token` parameter in https://github.com/spark/docs/commit/b7dbc7da25835ae5cacbe56e9ebfc91b1758b8ca. You could pass it as a string literal, but the most common case will be to save the token in a variable.

PR #541, opened 2017-01-11
Feature/faq eclipse debug

rickkas7:
FAQ for debugging using JTAG/SWD, Eclipse and OpenOCD


monkbroc:
Do you need a review since I wrote some OpenOCD guides too?

rickkas7:
Usually Will reviews the FAQs, but if you'd like to provide additional input that would be welcome as well.

brandoaire:
@monkbroc I think you or @technobly would be best to review this one since it's a pretty technical one!

technobly:
I think it would be useful for someone to actually go through it and follow it to see what snags come up.  I don't have the time this week though, but next week I could look at it!

monkbroc:
This guide is massive!

I reordered the top a bit. I put a link to the store for the Programmer Shield.

I'm not able to install the Eclipse plugins due to some Java SSL issues :( I think the cypher used by Sourceforge is not compatible with the latest JDK I have.

rickkas7:
It is indeed massive. It took a while to write, but it took forever to verify and take the screen shots because I did it on clean VM images of OS X, Windows 10 and Linux, and parts of it on Windows 7 as well. It was insane!

brandoaire:
Are we ready to merge? @monkbroc or @technobly  feel free to pull the trigger if you've read through it.

monkbroc:
I'm getting the same certificate exception when trying to install the GNU ARM Eclipse extension on Eclipse for Windows. I will try to update the Java version on my Windows machine.

image


I want to make sure to address this since I had the same error on 2 machines with a different OS.

rickkas7:
I'll re-run a test this weekend, if a resolution isn't found before then.

monkbroc:
I opened a GitHub issue with the GNU ARM Eclipse plugin. I tried Window and Linux with the latest JDK but I still couldn't get the plugin to install.
https://github.com/gnuarmeclipse/plug-ins/issues/171

rickkas7:
I see what you mean now. That definitely used to work, because I did it at least 5 times, but it fails for me now as well.

```
Unable to read repository at http://gnuarmeclipse.sourceforge.net/updates/content.xml.
Unable to read repository at http://gnuarmeclipse.sourceforge.net/updates/content.xml.
Received fatal alert: handshake_failure
```

I guess we'll have to wait until that's fixed.

monkbroc:
I would guess Sourceforge changed their SSL cert provider and Java doesn't trust the CA they used.

monkbroc:
I updated the steps to do a manual install of the plugin. It's not even more complicated.

I followed the steps until the end so we're good to go! :100: thanks for writing all of this, with screenshots! I know I'm going to use it :grinning:

PR #543, opened 2017-01-16
Add more info about disabling the preprocessor

pwallington:
Adding a mention of `else if (foo)` tripping up the preprocessor, and that Particle.h should be included manually after disabling it.

### DIFF
--- a/firmware/firmware.md
+++ b/firmware/firmware.md
@@ -11262,11 +11262,15 @@ make: *** [user] Error 2
 
 When you are using the Particle Cloud to compile your `.ino` source code, a preprocessor comes in to modify the code into C++ requirements before producing the binary file used to flash onto your devices.
 
-However, there might be instances where the preprocessor causes issues in your code. One example is the use of class/structs in your function parameters.
+However, there might be instances where the preprocessor causes issues in your code. Examples include:
 
+  * Using class/struct types in your function parameters (as above).
+  * Using single variables in conditions - e.g. `else if(foo)`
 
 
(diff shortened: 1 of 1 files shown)


firmware/firmware.md
```diff
@@ -11262,12 +11262,16 @@ make: *** [user] Error 2
 
 When you are using the Particle Cloud to compile your `.ino` source code, a preprocessor comes in to modify the code into C++ requirements before producing the binary file used to flash onto your devices.
 
-However, there might be instances where the preprocessor causes issues in your code. One example is the use of class/structs in your function parameters.
+However, there might be instances where the preprocessor causes issues in your code. Examples include:
 
+  * Using class/struct types in your function parameters (as above).
+  * Using single variables in conditions - e.g. `else if(foo)`
```
kennethlimcp, line 11268: hmm... interesting. Added an issue here: https://github.com/spark/buildpack-wiring-preprocessor/issues/6

I would prefer to fix the issue than listing as an issue in the preprocessor since that should work

pwallington: Agreed - I'd rather the preprocessor was fixed than documenting around it. However, preprocessor issues (including this one) have been discussed on the forums for a couple of years now, so I figured doc updates had a better chance of succeeding.

Additionally, there are mixed messages in the forum about including `Particle.h` or `application.h`, so I figured this should also be documented.

monkbroc:
I took a stab at finally fixing this issue. I've been bitten by it myself several times so it will be good for this to be gone.

https://github.com/spark/buildpack-wiring-preprocessor/pull/7

I'll go ahead and close this PR.

Thanks @pwallington and @kennethlimcp for pushing us to improve things!

PR #546, opened 2017-01-19
Fix documentation about libraries 2.0 in Build

stefan-moraru:
- Change 'Include in App' to 'Include in Project'.
- Update the 'Using libraries with Build' to point to the libraries section


stefan-moraru:
I will merge this as the release is today, and I double checked it locally.

PR #547, opened 2017-01-19
Update firmware.md

embedded-creations:
"All" is confusing when paired with "typically"

"All AT commands typically end" -> "AT commands typically end"


monkbroc:
Great. Thanks for the clarification.

PR #549, opened 2017-01-23
Incorrect link to localhost

dwaq:
I believe my modifications put the link in the right format to work on the website. I know the localhost part is definitely wrong though so a change is required regardless.


monkbroc:
Thanks!

PR #556, opened 2017-01-30
Add power pin limits callout to Photon datasheet.

harrisonhjones:
Added power pin limits to `power` section of Photon datasheet. 

@moors7 mentioned that we could also add it to the FAQ. I looked through it and I don't see a good section in the FAQ to add it to. Thoughts?

### DIFF
--- a/src/content/datasheets/photon-datasheet.md
+++ b/src/content/datasheets/photon-datasheet.md
@@ -58,9 +58,14 @@ The Photon comes in two physical forms: with headers and without. Prototyping is
 
 ### Power
 
-Power to the Photon is supplied via the on-board USB Micro B connector or directly via the VIN pin.  If power is supplied directly to the VIN pin, the voltage should be regulated between 3.6VDC and 5.5VDC.  When the Photon is powered via the USB port, VIN will output a voltage of approximately 4.8VDC due to a reverse polarity protection series schottky diode between V+ of USB and VIN. When used as an output, the max load on VIN is 1A.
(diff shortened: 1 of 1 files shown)


monkbroc:
I think the datasheet is sufficient.

monkbroc:
I queued this up in #19 so we include this in the next datasheet revision.

harrisonhjones:
@monkbroc Do you mean the datasheet **as is** is sufficient? or that you accept this PR?

monkbroc:
I meant that adding the info to the datasheet only, not the faq also, is sufficient. The PR is good 👍

I did not merge it so we can bundle all the datasheet changes together.

moors7:
Though it should suffice to put it in the datasheet, it definitely wouldn't hurt to put it in the FAQ as well, since it's not the first time that question (or a variation thereof) has come up. Perhaps a 'power section' is in order, highlighting the various 'specialties', current per pin, current for Vin/3.3V, tying Vbat to ... when not in use, etc.
Most/all the info is in the datasheets, but from the forum questions it's evident that those aren't always as easy to find. Just my 2 cents.

technobly:
I also think it's good to keep all of the numbers in one place as much as possible, i.e. the datasheet.  That way if we update those numbers we don't have to remember all of the other places where they were copied.  The FAQ could point to the datasheets for limits.

moors7:
Fair enough, it's a valid compromise. With that in mind, in this specific situation it seems unlikely those values will change (it's a hardware limitation?). With something like the LED patterns it makes more sense to keep those in one place, as opposed to the current three ;)

technobly:
Just updating the datasheets, and after considering this entry, it seems redundant since we have a Technical Specification section already with Absolute Maximum Ratings:
https://docs.particle.io/datasheets/photon-datasheet/#absolute-maximum-ratings

I'll add a link to that section in the Power section.

Thanks for the PR!

PR #557, opened 2017-01-30
Update for Librato Tags

benarent:
Librato recently introduced Tagged Metrics https://blog.librato.com/posts/tagged-metrics. This is turned on for all new Librato account. 

This updates to a new API endpoint and also removes default data that isn't valid data in Librato. https://www.librato.com/docs/api/#create-a-measurement.

Example response. 
image


adamyonk (approved):

monkbroc:
Great. Thanks for the update!

PR #590, opened 2017-03-16
Add link to VBAT resistor post in forum

embedded-creations:
It's been almost a year since instructions "will be posted shortly", might as well point to this rather than nothing


monkbroc:
Thanks! We'll include the full instructions in the next revision of the Electron datasheet (we version the datasheet so we do updates in batches). For now you're right that a link to the forum is better than nothing.

PR #591, opened 2017-03-17
Document product API endpoints

jme783:
Improve reference section of the API docs by adding a lot of endpoint coverage. Wait to merge until https://github.com/spark/api-node/pull/343 is merged and deployed


monkbroc:
I'm adding a paragraph and image about the product id and slug at the top because the slug is not described anywhere else in docs.
image

monkbroc:
I'll also change the heading of "Integrations" to "Integrations and webhooks" since searching for webhook no longer finds this section.

monkbroc (approved):

Issue #593, opened 2017-03-22
Doc page makes distinction between Photon/Electron but not WiFi/Cellular


rikonor:
Wanted to submit a PR for [mode-switching.md](https://github.com/spark/docs/blob/master/src/content/support/troubleshooting/mode-switching.md) where there are statements like:

> The default mode of the Electron is "automatic mode". This means that the Electron will attempt to connect to Wi-Fi automatically.

which are wrong since the Electron is connecting via a Cellular network, but the doc page is rendered using `{device}` and Wi-Fi is hard-coded.

kennethlimcp:
Thanks @rikonor! I'll fix that when i get on my laptop :)

kennethlimcp:
To be verified for electron:

> In this version of the code, when the {{device}} is plugged in, the LED will immediately start blinking. When a button attached to D0 is depressed (bringing DO to **LOW**), **Particle.connect()** will be called. If the {{device}} already has Wi-Fi credentials in memory, it will attempt to connect; otherwise, it will enter listening mode, and wait for your network name and password through the Particle mobile app or over USB.

Partial reword in this branch: https://github.com/spark/docs/tree/fix/593

kennethlimcp:
@technobly, some questions:

1. When a SIM is not inserted and we use `SEMI_AUTOMATIC`, does that go to listening mode or run the user code?

2. I would presume that if the user code runs and `Particle.connect()` is called (as per the example firmware) but a SIM is not present, it will enter listening mode?

This would help me with rewording the information above for the case of an electron.

I can test the behavior but that would involve me removing the SIM etc so I wanted to check in case you already know something. ;)

kennethlimcp:
1. When a SIM is not inserted and we use SEMI_AUTOMATIC, does that go to listening mode or run the user code?

The behavior is consistent where user code would run in `SEMI_AUTOMATIC` mode.

2. I would presume that if the user code runs and Particle.connect() is called (as per the example firmware) but a SIM is not present, it will enter listening mode?

Verified that the electron enters listening mode upon calling `Particle.connect()` if no SIM is present.

kennethlimcp:
Fixed ;)

PR #596, opened 2017-03-22
Fix/593

kennethlimcp:
Addresses: https://github.com/spark/docs/issues/593


monkbroc:
Thanks for the fix Kenneth. There's an even easier way to address this: use `{{network-type}}`. It puts cellular, Wi-Fi or network as appropriate. See https://github.com/spark/docs/blob/master/templates/helpers/network-type.js for the logic.

kennethlimcp:
@monkbroc thanks! fixed and rebased.

PR #597, opened 2017-03-22
Fix/examples

kennethlimcp:
Needs `fix/593` branch to be merged first as this is based on that branch


monkbroc:
Thanks for the fix Kenneth!

PR #598, opened 2017-03-24
Update client.read()

pamelafox:
I may be misreading it, but my guess is that it can't both put the data into `buffer` and report num read into `buffer`. So I think its meant to say `bytesRead`. Alternatively, it should probably just not say 'into...' at all, since variable name depends on usage.

### DIFF
--- a/firmware/firmware.md
+++ b/firmware/firmware.md
@@ -5969,7 +5969,7 @@ or `int read(uint8_t *buffer, size_t size)` reads all readily available bytes up
 bytesRead = client.read(buffer, length);
 ```
 
-Returns the number of bytes (or characters) read into `buffer`.
+Returns the number of bytes (or characters) read into `bytesRead`.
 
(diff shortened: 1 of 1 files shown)


bkobkobko:
Hi @pamelafox 

Your change is not correct.  The return value is indeed the number of bytes actually read, but they are put into the array called buffer (in C/C++ array arguments can be changed inside of functions).  So the text as written before was correct, the bytes are read into the array called buffer.

Hope this makes sense.

pamelafox:
Ohh! I was confused by the ambiguity of the sentence wording. I read it as "it returns the number (...) into buffer". I'd suggest rewording to "Returns the number of bytes (or characters) read." I suppose you can leave it as is if you're trying to make sure people realize which usage you're referring to.

kennethlimcp:
Closing since it has been explained by @bkobkobko!

PR #599, opened 2017-03-24
Give fix-it tip for Cloud not connected

pamelafox:
Background: Our code uses TCPClient, and in its current stage of development, it often goes into the GREEN stage, and we need to put it into safe mode to flash the new code. We were very perplexed when it first happened, and nearly ordered a new Photon as we didn't understand why it wasn't cloud-connected. We figured it out after some forum hunting though.

So my proposal is to put a tip in the device modes doc, based off the most frequent tip in the forum.
Y'all likely know more about other causes of Cloud not being connected, so you may have better wording.

Thanks!

### DIFF
--- a/src/content/guide/getting-started/modes.md
+++ b/src/content/guide/getting-started/modes.md
@@ -333,6 +333,8 @@ If the Wi-Fi module is on but not connected to a network, your {{device}} will b
 
 When the {{device}} is connected to a {{#if electron}}cellular{{/if}}{{#if photon}}Wi-Fi{{/if}}{{#if core}}Wi-Fi{{/if}} network but not to the cloud, it will breathe green.
 
+Sometimes that happens due to flashed code that interferes with its ability to connect to the cloud. You can try going into Safe Mode with the instructions above, and then flash new code.
+
 
 ### Bad Public Key
 


kennethlimcp:
LGTM. I can reword and add a link to `safe mode` once this is pulled in.

ScruffR:
I now have already updated the docs (based on your experience) with this:

> This can be caused by the currently running application firmware which may interfere with the cloud maintenance tasks which is usually executed between iterations of `loop()` or via an explicit call of 
> [`Particle.process()`](/reference/firmware/photon/#particle-process-). That commonly happens when the code blocks for more than 10 seconds. 
> In addition to regularly allowing for cloud maintenance (via dropping out of `loop()` and/or calling 
> `Particle.process()`) you can take manual control of the [connection](/reference/firmware/photon/#cloud-functions), choose an better suited [`SYSTEM_MODE`](/reference/firmware/photon/#system-modes) and/or opt for [`SYSTEM_THREAD(ENABLED)`](/reference/firmware/photon/#system-thread).
> To correct the "offending" firmware you may need to flash new firmware either via USB or [Safe Mode](#safe-mode).
> 
https://docs.particle.io/guide/getting-started/modes/photon/#cloud-not-connected

I hope this tells it all ;-)

kennethlimcp:
@pamelafox thanks for your inputs! Appreciate it and keep em' coming ;)

pamelafox:
Oh, fantastic explanation, thank you! I had a TODO to research using MANUAL mode and Particle.process, and that explanation gives me great pointers.

pamelafox:
Just reading #system-thread now, I think there's a typo as there's a dangling clause masquerading as a full sentence: ". while AUTOMATIC mode connects to the cloud as soon as possible."

https://docs.particle.io/reference/firmware/photon/#system-thread

kennethlimcp:
fixed: https://github.com/spark/docs/commit/e961be0ec644b43a36bd2e338d6df380e67da96a

PR #601, opened 2017-03-26
Update firmware.md

hpssjellis:
As per comments at https://community.particle.io/t/class-set-of-accelerometers/30302/17 updating the docs so that a float can be easily formatted to x number of decimal places when converted to a string


kennethlimcp:
Verified to be available at: https://github.com/spark/firmware/blob/develop/wiring/src/spark_wiring_string.cpp#L149

@hpssjellis thanks!

PR #605, opened 2017-03-31
Update GCC-ARM toolchain to point to exe installer

badmotorfinger:
Link was pointing to a zip, not an exe, making the instructions confusing.


kennethlimcp:
LGTM. @vincpa thanks!

PR #618, opened 2017-04-16
Update start.md

zhengs:
The link to the iPhone app is wrong. It should have been the "Particle - Build IoT projects - WiFi or Cellular" app instead of "Particle Core", which seems to be the old app for Core, otherwise the connection to the Photon device would always fail. It took me 1 hour to figure this out.


kennethlimcp:
@zhengs sorry for the confusion and thanks for the PR!!

PR #621, opened 2017-04-25
device information, event subscription method name fix

CityVibes:
Update on cloud sdk part of Android documentation:
1) Adding section for reading device information.
2) Fixed subscription method name while calling subscription on ParticleDevice.


monkbroc:
Awesome! Thanks for the update.

PR #622, opened 2017-05-02
Updated the first paragraph of "Particle.connect()" function

indie12345:
First paragraph of Particle.connect() for electron device was saying:
"Particle.connect() connects the device to the Cloud. This will automatically activate the Wi-Fi module and attempt to connect to a Wi-Fi network if the device is not already connected to a network."

I replaced the mention of WiFi to cellular connection.

### DIFF
--- a/firmware/firmware.md
+++ b/firmware/firmware.md
@@ -460,7 +460,7 @@ Particle.unsubscribe();
 
 ### Particle.connect()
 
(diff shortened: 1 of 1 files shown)


monkbroc:
Awesome! Thanks for that. I updated your change to use the `{{network-type}}` macro so it will show Wi-Fi for the Photon and cellular for the Electron.

PR #626, opened 2017-05-12
Add docs for setting up a Photon using the web app

stefan-moraru:


brandoaire:
Hey @stefan-moraru I don't want to block on this one and don't think I'll have the chance to review before it needs to go live. Can you pull in someone else for copy editing? Also, not sure we should change the docs to point to setup.particle.io as our primary setup until we've:

- [  ] Added the info cards to the end of setup
- [  ] Added compatibility for Internet Explorer

kennethlimcp:
@stefan-moraru should this be merged since the announcement is already out in the wild?

monkbroc:
I went ahead and merged. Thanks for the reminder @kennethlimcp

brandoaire:
@monkbroc @stefan-moraru as far as I know my concerns from above still haven't been addressed so I'm surprised to see this merged in. I see lack of compatibility with IE as a blocker to making browser photon setup the default configuration flow, especially since as far as I know we're not doing proactive alerting to prevent people using IE from visiting the site.

My suggestion is to revert this merge until after Web based Photon setup is ready for beta and has received a formal approval from the product team, as I noted previously.

PR #631, opened 2017-05-24
Update LED blink code description

AlexSneedMiller:
The "Flashing cyan" blink sequence happens when you are connected to a Wifi network (i.e. have an IP), but have not yet connected to the internet. I'm not sure if this is 100% correct or if there are other scenarios where this blink sequence will happen, but that is what I have observed with the Photon.

This might be redundant with the Flash red twice error code, but I'm not sure how long it takes for that error code to appear. Perhaps adding a description of a timeout will clear that up?


kennethlimcp:
That's correct! You will see flashing green which means connecting to WiFi network which transitions to flashing cyan which means connecting to ☁️ . We can always pull in your description if it makes things clearer.

kennethlimcp:
LGTM.

Just wanted a second opinion @pkourany

AlexSneedMiller:
do you want to merge this or should we close it?

kennethlimcp:
Thanks @AlexSneedMiller! Merged now ;)

monkbroc:
Thanks for reminding us about this and thanks for your contribution!

PR #632, opened 2017-05-24
Move firmware reference back and remove subtree

monkbroc:
Having the firmware docs in a subtree shared between the docs and firmware repos was leading to more conflicts before firmware releases leading to unplanned work. Let's go back to making PRs directly in the docs repo for new firmware docs.


m-mcgowan (approved):
ok.  Although as some background research I'd like to understand why this isn't working for us, since it's worked well for me in other projects.

technobly:
FYI, to remove the subtree from local `spark/docs` use `git remote rm firmware-docs`

PR #640, opened 2017-06-04
The return value of Time.now() is unsigned.

rvnash:


kennethlimcp:
@rvnash how did you verify this? `now()` returns a value of type `time_t` but the signed and unsigned property needs to be determined based on our toolchain

rvnash:
I apologize but the firmware is a bit confusing.

In hal/src/photon/wiced/network/NetX/ver5.5_sp1/netx_bsd_layer/nx_bsd.h there is:
typedef ULONG       time_t;

and elsewhere in platform/NET/CC3000/CC3000_Host_Driver/cc3000_common.h it is defined as:
typedef INT32 time_t;

But neither of these appears to be the real definition. The real definition appears to be buried in the compilers include paths and appears to be found in 
arm-none-eabi/include/machine/types.h 
#define	_TIME_T_	long
arm-none-eabi/include/sys/types.h:
typedef _TIME_T_ time_t;

It is signed and a long, which in ARM is equivalent to int. So the documentation is correct.

PR #642, opened 2017-06-15
Add example use cases

sideye:
Taken from "extended use cases" in the Azure IoT hub documentation, under "get started" subsection: https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-get-started


monkbroc:
Great! Thanks!

PR #649, opened 2017-06-30
Replace "breathe <color>" with "breathing <color>"

embedded-creations:
"breathing \<color\>" seems to be the most popular way to describe breathing LED behavior on a Particle device.

"electron "breathe cyan"" has 166 results on Google
"electron "breathing cyan"" has 1670 results on Google

However, in this guide to the LEDs, "breathe \<color\>" is used a few times, so I can't find "breathing blue" for example, I think it's not documented, and I go to the forum to search for "breathing blue" instead.  I replaced "breathe \<color\>" with "breathing \<color\>"

There's probably the same problem with something like "blinks \<color\>" instead of "blinking \<color\>", but I didn't look for that.


monkbroc:
Thanks for updating the terminology!

PR #650, opened 2017-07-01
Hints provided by ScruffR that should be in the documentation.

jr3cermak:
REF:

[Nested header structure](https://community.particle.io/t/porting-wolfssl-library/34131/6?u=cermak): I've confirmed that this works.

[use of return possible in loop() and/or setup()](, this can be taken for granted.  I didn't think this was permitted in these special functions, so I think it should be suggested that it can be used to pre-empt full runs of these functions.


kennethlimcp:
@jr3cermak, thanks for the inputs!

PR #660, opened 2017-07-17
Let's not joke on gender ;)

Alain1405:


monkbroc:
Right on. Thanks for the correction.

PR #665, opened 2017-07-22
Adding links for circuit.io and sunstone.com

rdpoor:
Added links for circuit.io (schematic capture and PC layout) and Sunstone Circuits (quick-turn online PC fabrication).  I've used both together and been very happy with the results.


technobly:
Thanks for the suggested edit.  For this list though it is very subjective based on what you are comfortable with and who's giving you good deals at the moment.  We should keep this list short, based on what Particle has used internally, otherwise it can start to get wild with endorsements :)

FYI: Sunstone has been around for 40 years ^_^

PR #666, opened 2017-07-27
p1-datasheet-v008, photon-datasheet-v015 and electron-datasheet-v005

technobly:
All open errata [here](https://github.com/spark/docs/issues/19) incorporated into the next revision of datasheet for Photon and P1


mohitbhoite (approved):
Love the new block diagrams and clarity on the power supply requirements.
On a side note, for the landing pattern of the Photon (headerless), it is not immediately clear which one is the top view and which is the bottom view.

mohitbhoite (approved):
Looks good @technobly

PR #678, opened 2017-08-30
Improved documentation

jenschr:
There's no details currently on the 1Mb Flash chip in the P1 modules. Adding this tiny information will save others from decapping a chip (like I did) to learn the name of the chip so you can locate a datasheet.


monkbroc:
Thanks! We'll include that change as an errata to the datasheet.

technobly (approved):

PR #685, opened 2017-09-19
Update firmware.md

sean-boyd:
Line 5528 reads "message.length" instead of "message.len".


kennethlimcp:
LGTM! @seanmatthewboyd thanks :)

PR #686, opened 2017-09-19
Keep titles in lowercase in the reference section

monkbroc:
Several pages in the reference section have headers that are commands (CLI) or function names (JS client). Transforming them to uppercase makes it harder for users to find the correct capitalization.

Also switch the section id to the lowercase filename because it was generating stuff like `id="Command Line Interface (CLI)-content"` otherwise.

Example:
Header for https://docs.particle.io/reference/javascript/#createuser should say createUser not CREATEUSER


anatoly-particle:
Cool PR.

anatoly-particle (commented):
Good work!

PR #688, opened 2017-09-26
Fix top bar layout on mobile

anatoly-particle:
Fixes topbar for small screens _`<= 452 px`_ by hiding the image and reducing margins.

Variable is used for consistency with other code.

### DIFF
--- a/src/assets/less/header.less
+++ b/src/assets/less/header.less
@@ -274,3 +274,17 @@ a.nav-device {
     display: none;
   }
 }
+
+@media (max-width: @small-size) { // 452px to be precise
+  .header .left {
+    > a {
+      display: none;
+    }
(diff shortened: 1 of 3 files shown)


monkbroc (approved):
I added a commit to show the logo on the home page of the mobile docs, otherwise the navbar at the top was empty.

monkbroc:
If you want to check the home page of docs, open http://localhost:8080 in incognito mode.

If you have a better approach that using a `:not(.redirector-header)` pseudo-class, please make the change. If you're OK with it, you can merge the PR and it will get deployed!

anatoly-particle:
Aha, thanks, haven't thought about `redirector`.
Good catch! 👍🏼

PR #689, opened 2017-09-26
Add partial templates to Metalsmith watcher

anatoly-particle:
Glob to trigger rebuilds `content/**/*.md` may be too loose, though.


monkbroc (approved):
Looks good. Merge it!

PR #691, opened 2017-10-12
Add instructions how to use Serial port

anatoly-particle:
I couldn't find the instructions how to use Serial port, so I copied text from “Troubleshooting Tools: Listening Mode Commands“.


monkbroc (commented):
I'm not sure if Windows users still need to install drivers. I asked Sergey in the firmware team for his opinion.

beriberikix:
Windows *generally* needs [Zadig](http://zadig.akeo.ie/) installed for USB 2 Serial. Might be worth testing against this, since it is broadly used and supported.

monkbroc:
OK. Let's the instructions like this since Andrey mentioned that > 0.6.0 firmware doesn't need Zadig.

PR #693, opened 2017-10-18
Add PDF generation for datasheets

anatoly-particle:


scripts/pdf-generation/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+To generate PDF from all the datasheets just run (it should work on macOS and Linux at least):
+
+```
+npm run pdf-generation
+```
+
+**But make sure that you've done following things before:**
+- Install all the **dependencies** with `npm install`
+- Install **wkhtmltopdf** tool for your system from https://wkhtmltopdf.org/downloads.html
```
monkbroc: What version do you use? We should document that.

monkbroc (approved):
:+1:

PR #694, opened 2017-10-22
correctly print object in getEventStream example

chrisgervang:
consoles print "[object Object]" instead of the full object when using "+" because it tries to cast to a string. Using a comma instead prints the full object in browsers and most terminals.


monkbroc:
Great! Thanks.

PR #695, opened 2017-10-24
Compress all images in \src\assets\images directory 

dwaq:
I know this is a huge pull request, but I thought it would help not only viewers of your website, but anyone else that may want to make changes to the repo (cloning it now takes a while). Compressing every `.png` and `.jpg` file in the `\src\assets\images` directory saved a ton of space.

|Location|Original Size (MB)|New Size (MB)|Size Reduction (MB)|Compression|
|-|-|-|-|-|
|\src\assets\images|88.8|49.3|39.5|56%|
|Without .git directory|262|222|40|85%|
|Entire Repo|785|769|16|98%|

You can see that it didn't save much space over the entire repo because the git history is now inflated. I believe this can be improved by rewriting the history, but this is not an easy change.

I think a lot of the remaining space is taken by `.pdf`s. If this is also something that is desired, I can compress those as well.


monkbroc:
That's cool! Thanks for taking the time to do this.

I'm not that concerned with the size of the repo but since the images actually have to be loaded by each person viewing the docs, compressing those make a huge difference!

anatoly-particle:
Hi, @dwaq! Thank you for your work and the PR!

A couple of questions:
1) Why were two images added?
While almost all files are modified, two images were added: `mki-dc-motor.jpg`, `mki-pushbutton.jpg`.
2) Was the compression lossless or not?
I've noticed you used **optmimizilla** tool.
Btw, have you used it with the CLI or web-interface?

dwaq:
1. Those images must be a mistake. They might belong somewhere else and I could've placed them in the wrong location. I looked to see if I made this mistake but must've missed those two. 
2. The compression was lossy but I was hard pressed to tell the difference. I used the web interface and some scripting. I already got too far before I realized just how many images were in this repo.

dwaq:
@anatoly-particle I fixed the two extra images issue in the latest commit. As I mentioned, I had a script that would remove the "-min" substring from the optimizilla-generated image. Those two images were called mk-mini-xxx.jpg, so when I removed the -min, it changed the filename to mki-xxx.jpg.

anatoly-particle:
@dwaq Aha, thanks, that makes sense.

If you don't mind, I will test the lossy conversion with **ImageOptim** and compare the results.
So far `images/` after losslees compression is 74.3 MB, yours is 51.7 MB, the original is 93.1 MB.

Also, I will need to check whether all images are not broken after the compression.

anatoly-particle:
Hi @dwaq, thank you again for your suggestion and your work!

Can you please take a look at PR #698, where I do the controlled lossy compression and got similar but just slightly better results.

dwaq

I would like to add the commit on top of yours, but, as you rightly said, it will inflate the git history even more.

If you don't mind, I will merge that (#698) request (where you are acknowledged for your work).

dwaq:
Yes, go ahead and use yours. I was just trying to find a useful project to complete my Hacktoberfest PR's  :smiling_imp:.
As mentioned earlier in this thread, the reduced image size is helpful for people viewing the documentation on the website, but the huge size of this repo may scare people away from contributing. It might be useful to find other ways to reduce the total size of the repo.

anatoly-particle:
Ha! Good luck! 😃
@dwaq How many of four PRs you have so far?

Yes, you absolutely right about the repo size.
Please let me know if I can somehow help you with the Hacktoberfest in this (Docs) repo.

dwaq:
@anatoly-particle Thanks! I actually finished all 4. I had 2 others that I wanted to work on in this repo, but I don't think I'll have the time now. I'll add them as issues so maybe others can work on them. If you add the `hacktoberfest` label to the issue, they will be publicly searchable and maybe another newcomer will pick them up.

PR #697, opened 2017-10-25
Make page numbers links in datasheet TOC

monkbroc:
I was under the impression there were no links in the table of contents for the PDF datasheets because I always clicked on the page numbers, maybe because they were blue instead of black. I just saw the titles were links so I added a transformation to make the page numbers links as well.

I also made it so the PDFs are not deleted before being regenerated. My PDF viewer will automatically refresh when the document is overwritten which is quite convenient.


anatoly-particle:
Nice catch on the page numbers links. 👍🏼

PR #705, opened 2017-10-30
Update firmware.md

joergseiler:
Added note that published event names with spaces don't allow event subscriptions to function as anticipated.


monkbroc:
Thanks for clarifying the docs for Particle.publish and congrats on your first GitHub PR!

PR #706, opened 2017-11-04
Added documentation for UDP.receivePacket()

Walter-Stroebel:
This function was not documented.


kennethlimcp:
@technobly @pkourany any comments?

PR #708, opened 2017-11-08
Added E series documentation and reorganized datasheets

mohitbhoite:
@monkbroc @anatoly-particle Please take a look. 
I’ll add assets for PDF version of the datasheet later.


monkbroc (approved):
With Anatoly and I crunching to finish the E Series web setup before modules ship, I don't think we can give thorough review. We talked yesterday about the approach so if it builds, ship it! :shipit:

PR #712, opened 2017-12-01
CLI installer

monkbroc:
Update instructions to install the CLI from the installer.

The main page that changed is https://docs.staging.particle.io/guide/tools-and-features/cli/photon/


rickkas7:
Looks good to me!

PR #713, opened 2017-12-05
Adding a reference to the ember-particle ember addon.

mileszim:
Nothing more, nothing less


mileszim:
Looks like the build process is not a fan of anchor links, as it's 404ing each one from README.md. What's the process for fixing this?

monkbroc:
Thanks for this. If the Travis build fails on `master` I'll take care of it.

PR #722, opened 2018-01-10
Revert "I2C, Serial, Photon Serial FAQs"

rickkas7:
Reverts particle-iot/docs#721


rickkas7:
Wasn't necessary to revert, failure of original commit was caused by something else

PR #725, opened 2018-01-24
Fix broken img tags

ghost:


monkbroc:
Thank you for the fix!

PR #727, opened 2018-01-26
FindStrongestSSID example always returns null (logic error)

petermm:
strongest_rssi is initialised as 0

the logic is: if ((ap.rssi < 0) && (ap.rssi > strongest_rssi))

thus the strongestFinder.scan() example always returns null as rssi values are between -127 and -1 - and never satisfying (ap.rssi > strongest_rssi)

this PR changes the init value to -128 thus any ap found (including with -127 rssi) will be matched - and the strongest will be found accordingly.

### DIFF
--- a/src/content/reference/firmware.md
+++ b/src/content/reference/firmware.md
@@ -1277,7 +1277,7 @@ public:
     const char* scan()
     {
         // initialize data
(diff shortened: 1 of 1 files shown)


monkbroc:
Thanks!

PR #731, opened 2018-01-31
InfluxDB integration

davidgs:
Docs for using the InfluxDB Webhook


rickkas7:
This is great! If you could fix the conflict in the .md file and update from master I'll go ahead and merge once we get a clean build.

```
+<<<<<<< Local Changes
 +Particle has teamed up with InfluxData to create anplatform easy webhook integration
 +=======
 +Particle has teamed up with InfluxData to create a an easy webhook integration
 +>>>>>>> External Changes
```

Thank you for the tutorial!

PR #734, opened 2018-02-12
dfu-util-0.8 at amazonaws is aging and 32-bit

csylvain:
installing dfu-util-0.9 and later from distro repository is better advice than grabbing v0.8 from an old spark-assets repo. i needed a 64-bit binary.


technobly:
Thanks!  I wanted to make a slight change to your contribution so I committed it directly here: https://github.com/particle-iot/docs/commit/d19a6bcd7440f4ae7152fd235e5e59039ccd4029

csylvain:
that's fine. every step is progress!

PR #739, opened 2018-02-23
update for "Orientation Awareness"

csylvain:
example 6_Orientation does not visibly illuminate the lowest LED. it gets turned off faster than can be seen by eye. this revision also publishes the lowest LED position so the user can monitor and verify Internet Button orientation from Particle console.


monkbroc:
Same code as https://github.com/particle-iot/InternetButton/pull/30
Thanks for updating this one too.

PR #743, opened 2018-02-23
Documentation for attachInterruptDirect() / detachInterruptDirect()

avtolstoy:
References:
- particle-iot/firmware#1394


ScruffR:
An example of how to mimik a pin interrupt - including edge detection - might be good for the docs.
Also a link to the pin  map and a description how to get from the GPIO "name" to the respective nIRQ would help.

PR #754, opened 2018-03-14
Change heading fonts

anatoly-particle:
Preview → https://docs.staging.particle.io

### DIFF
--- a/src/assets/less/content.less
+++ b/src/assets/less/content.less
@@ -379,6 +379,11 @@ body:not(.big-h3) .content.reference h3 {
   font-family: @body-font;
   font-size: 16px;
 }
+.content.reference h2 {
+  text-transform: none;
+  font-family: @body-font;
+  font-size: 22px;
+}
 .content h1:first-child {
   padding: 0;
 }
--- a/src/assets/less/content.less
+++ b/src/assets/less/content.less
@@ -358,9 +358,9 @@ body.big-h3 .content h3:before {
 body:not(.big-h3) .content h3 {
   color: @dark-gray;
   text-transform: uppercase;
-  font-size: 16px;
+  font-size: 18px;
   letter-spacing: 1px;
-  font-family: @body-font;
+  font-family: @header-font;
   padding-top: 30px;
   padding-bottom: 10px;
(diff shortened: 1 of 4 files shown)


jme783:
Hey @anatoly-particle, this is almost there. Nice work! What I'm unclear on is why headers are treated differently in some parts of the reference section compared to the guide, FAQ, etc. 

I actually really like how the fonts look when the _only_ use of DIN on the page content are the page titles (`h1`s), while the rest are Avenier. Essentially what I'm suggesting is using the styling on the firmware reference pages across the docs. The most important thing is consistency.

Like this, everywhere:
screen shot 2018-03-14 at 7 11 32 am

Something like:

```
h2 {
  font-family: AvenirNext,sans-serif;
  text-transform: none;
  font-size: 24px;
}

h3 {
  font-family: AvenirNext,sans-serif;
  text-transform: none;
  font-size: 18px;
  letter-spacing: 0;
}

jme783 (changes requested):
See above comment

anatoly-particle:
@jme783, thanks again for the feedback
Here is a small fix to make changes you suggested.
`!important` is a hacky way to do it, but with current styles it is the easiest way to enofrce that styling.
One thing to note is that headings on the sidebar change as well. I like it more:

Before | After
--- | ---
before | after

Changes on staging → https://docs.staging.particle.io/faq/particle-devices/led-troubleshooting/photon/

jme783 (approved):

jme783:
This looks good. I'd say merge. I do think the icons in the sidebar are still too big. Might be worth a separate small fix:

screen shot 2018-03-15 at 6 33 24 am

screen shot 2018-03-15 at 6 33 41 am

anatoly-particle:
@jme783, I've fixed the icons.
Can you please review changes again?

Before | After
--- | ---
before-icon | after-icon
before-list | after-list

Preview on staging → https://docs.staging.particle.io/reference/javascript/#device-management

monkbroc:
Since you have taken into account all Jeff's feedback, there's no need to wait for an additional review. Let's ship this now.

PR #756, opened 2018-03-23
Update webhook reference section to discuss new error recovery algorithm

jme783:
Also add some docs on webhook audit logs, which we have been missing for a while!


monkbroc (approved):
:+1: The Good Internet Citizenship Office would approve too.

PR #760, opened 2018-04-06
[Firmware.md] Document 3rd-party SIM issues

kubark42:
When multi-threading, a situation arises with `keepAlive()`. This is documented at https://github.com/particle-iot/firmware/issues/1482, and the PR updates the documentation to reflect this.


technobly:
Thanks @kubark42. We don't typically add firmware issues to the docs though, as they are a nightmare to find and pull back out once the issue issue is solved.  I'll leave your issue here as a way people can search for the problem, but it will be closed.

PR #761, opened 2018-04-09
[Cell]: `command()` must follow `on()`

kubark42:
As per https://community.particle.io/t/cellular-command-always-returns-1-on-electron/20406/16, when using SEMI_AUTOMATIC or MANUAL system modes, `Cellular.on()` must be called prior to `Cellular.command()`. This documents that nuance.


ScruffR:
I added the alternatives `Cellular.connect() and `Particle.connect()` and linked to the respective docs sections too.

PR #763, opened 2018-04-12
Use more triple braces in webhook examples

monkbroc:
Use more triple braces in webhook examples to avoid people accidentally HTML encoding their webhook data


dmiddlecamp:
Looks awesome, thank you!

dmiddlecamp (approved):
Looks awesome, thank you!

PR #781, opened 2018-06-05
Fix Wi-Fi mention for Electron `Particle.process()`

bsatrom:
Quick fix to the Electron docs for Particle.process

### DIFF
--- a/src/content/reference/firmware.md
+++ b/src/content/reference/firmware.md
@@ -604,14 +604,18 @@ The keep alive duration varies by mobile network operator. The default keepalive
 Runs the background loop. This is the public API for the former internal function
 `SPARK_WLAN_Loop()`.
 
+{{#if has-wifi}}
 `Particle.process()` checks the Wi-Fi module for incoming messages from the Cloud,
 and processes any messages that have come in. It also sends keep-alive pings to the Cloud,
 so if it's not called frequently, the connection to the Cloud may be lost.
 
-{{#if has-wifi}}
 Even in non-cloud-bound applications it can still be advisable to call `Particle.process()` to explicitly provide some processor time to the Wi-Fi module (e.g. immediately after `WiFi.ready()` to update system variables).
 {{/if}}
 {{#if has-cellular}}
+`Particle.process()` checks the Cellular module for incoming messages from the Cloud,
+and processes any messages that have come in. It also sends keep-alive pings to the Cloud,
+so if it's not called frequently, the connection to the Cloud may be lost.
+
 Even in non-cloud-bound applications it can still be advisable to call `Particle.process()` to explicitly provide some processor time to the Cellular module (e.g. immediately after `Cellular.ready()` to update system variables).
 {{/if}}
 


anatoly-particle (approved):
👍

monkbroc:
To avoid duplication when the only difference is the type of network, use the `{{network-type}}` helper.

See definition in https://github.com/particle-iot/docs/blob/41aaaa1839e27dae6f0bf54427f2a30dd2be0e3f/templates/helpers/network-type.js and usage in https://github.com/particle-iot/docs/blob/7cfbd748324c71b8374690788cf5a6691cdf9054/src/content/support/troubleshooting/troubleshooting-support.md

@bsatrom Do you want to open another PR with `{{network-type}}` replacing Wi-Fi in the original commit?

Other helpful helpers are `{{device}}`, `{{a-device}}`, `{{reset-button}}`, `{{system-button}}` and the `{{if has-<feature>}}` construct where `<feature>` can be any of the strings in https://github.com/particle-iot/docs/blob/41aaaa1839e27dae6f0bf54427f2a30dd2be0e3f/config/device_features.json

anatoly-particle:
Very useful! Thanks, @monkbroc!

Side note: `network-type` helper will return lowercased **_“cellular”_**.

bsatrom:
Thanks @monkbroc, will do!

monkbroc:
> Side note: network-type helper will return lowercased “cellular”.

If that's a problem you can always define a new helper called `titltecase` similar to https://github.com/particle-iot/docs/blob/41aaaa1839e27dae6f0bf54427f2a30dd2be0e3f/templates/helpers/titleize.js and use it like `{{titlecase network-type}}`

bsatrom:
To @anatoly-particle's point, what can we do in the case where the docs are referring to the WiFi or Cellular classes? Is there a helper for these as well?

monkbroc:
Make a new helper :)

bsatrom:
that's what I was just wondering. :D Will do.

monkbroc:
Actually do you have specific instances of referring to WiFi or Cellular class? That sounds rare enough that it might be easier to just do `{{if has-wifi}}` or `{{if has-cellular}}`

bsatrom:
`Particle.process()` was one place I came across it: https://docs.particle.io/reference/firmware/electron/#particle-process-

> Even in non-cloud-bound applications it can still be advisable to call Particle.process() to explicitly provide some processor time to the Cellular module (e.g. immediately after Cellular.ready() to update system variables).

monkbroc:
If it's just one case, then I'd do this:

```
Even in non-cloud-bound applications it can still be advisable to call Particle.process() to explicitly provide some processor time to the {{network-type}} module (e.g. immediately after {{if has-wifi}}WiFi{{else}}{{if has-cellular}}Cellular{{/if}}{{/if}}.ready() to update system variables).
```

It's not generalized, but it's also hard to know if it will apply for future boards that won't have Wi-Fi or cellular.

bsatrom:
Thanks Julien, will do

PR #782, opened 2018-06-12
Clean up based on comments from #781

bsatrom:


anatoly-particle (approved):

anatoly-particle:
@bsatrom 👍 

---

**Update 1:** Slightly fixed the usage of `{{#if …}}` template.

---

**Update 2:** Those two paragraphs are visible when Raspberry Pi is chosen (which is probably unintended).
So, wrapped them with `{{#unless has-linux}}` template, as it is the only distinctive device feature for Raspberry Pi.
https://github.com/particle-iot/docs/blob/ae53c69e767abaed2553686babdffc90eb719f4f/config/device_features.json#L90

_Context:_ these device features are being added here https://github.com/particle-iot/docs/blob/ae53c69e767abaed2553686babdffc90eb719f4f/scripts/device_feature_flags.js#L25

---

cc @monkbroc

bsatrom:
Awesome, thanks for the update and context, @anatoly-particle!

PR #788, opened 2018-07-03
Add a guide page for the Console Event Logs page

stefan-moraru:


monkbroc (commented):
This should replace the "Logs" section of the Console guide, not create a new page. https://docs.particle.io/guide/tools-and-features/console/#logs

Avoid talking about how this is an improvement over something that came before. The docs talk about how things work now and any reference to previous implementations are going to get stale quickly.

The voice of the docs is also a bit off. Avoid overt praise of the tech like "Filters are a great way to narrow down your search". You can say "Filters let you narrow down your search". You can remove "Because the page was developed with scalability in mind it is now able to handle hundreds of events coming from your devices at the same time." As a user I expect the event log to function properly no matter where I am in the product cycle. I don't need to be reassured that the event log is going to work. An example that's ok is "The new Event Logs page is tailored around improving this experience. It provides a handful of tools, like filtering, modifiers which enable you to narrow down your search for events, making it easier to view only the data that is relevant to you." since it highlights the customer value of filtering.

src/content/guide/tools-and-features/event-logs.md
```diff
@@ -0,0 +1,69 @@
...
+You can filter the events by writing anything in the input. Your query will be compared with the event name, data, publisher, and date.
+<img src="/assets/images/eventlogs/filter_word.png" class="full-width" />
+
+#### Modifiers
+
+Besides writing in the input, you can use modifiers to narrow down your search even further.
+
+<img src="/assets/images/eventlogs/filters.png" class="full-width" />
+
+- `product` Filter by product (example: `product:my-product`)
```
monkbroc: I thought the `product` filter was going away. It doesn't make sense to me: if I'm on a user or individual device page, filtering by product doesn't make sense. If I'm on a product page, it should always filter by product anyway.

PR #805, opened 2018-08-08
Consolidate status LED information

rickkas7:
Move less common modes from getting started to a new reference section,
remove the rarely used page in support.


monkbroc:
Fantastic! Do you know the exact patterns described in https://docs.particle.io/reference/led/photon/#red-blink-basic-errors ? It would be helpful to add animations to distinguish between these since they are rare and fast.

PR #813, opened 2018-08-14
Safe times to register functions and variables

rickkas7:
Also a little more text about when to use functions, variables, publish, and subscribe.


mprogers:
That was awesomely fast work, thank you!!

PR #835, opened 2018-09-11
Correct spelling in Particle publish firmware reference

remear:
I believe `connecvtion` was meant to be `connection`.


rickkas7:
Thank you!

PR #855, opened 2018-10-15
Possible typo?

3leftturns:
Heading says "Browser" but detail text indicates to install using the JS Bower package management tool. Is this a typo?


monkbroc:
No, this just means that if you want to use particle-api-js in the browser, you can use a packaging tool like bower or load the module directly from a CDN.

PR #858, opened 2018-10-26
Update xenon.md

dhhagan:
Updated the link to the Xenon datasheet which was previously sending you to the hardware repo


rickkas7:
Thanks! I'm going to include this change with the changes to fix the build error.

Issue #867, opened 2018-10-30
[Boron] Cellular.setActiveSim() not documented


kennethlimcp:
This will be needed for people using 3rd party SIM though everyone will probably start off with the eSIM when first unboxing it.

rickkas7:
Fixed via: https://github.com/particle-iot/docs/pull/869

Issue #864, opened 2018-10-29
Boron reference is missing Cellular section


kennethlimcp:
Stuff like switching to external SIM and setting APN (also that APN is now stored in NVM) are not yet available

rickkas7:
Fixed via: https://github.com/particle-iot/docs/pull/869

PR #869, opened 2018-10-30
Document Cellular setCredentials and setActiveSim

rickkas7:
- Rename sections in device menu for Photon / P Series and Electron / E Series
- Document Cellular.setCredentials and Cellular.setActiveSim (#864 #867)

### DIFF
--- a/config/device_features.json
+++ b/config/device_features.json
@@ -167,7 +167,9 @@
     "button-mirror",
     "rgb-mirror",
     "spi-settings",
-    "tcpclient"	  
+    "tcpclient",
+    "set-credentials",
+    "set-active-sim"
   ],
   "Xenon": [
 	"nrf52",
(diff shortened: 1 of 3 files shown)


technobly:
@rickkas7 it is work mentioning that on Boron LTE devices with external SIMs, they may work by default without setting the credentials.   Some LTE SIM cards have the APN details embedded internally, and the Device OS will try to automatically use those if the MCC and MNC do not match one of the cards Particle uses. If the MCC/MNC does overlap with a Particle SIM or the card does not support automatic APN usage, the external SIM credentials would need to be explicitly specified via Cellular.setCredentials. `External SIM` does also need to be selected though with Cellular.setActiveSim or via the bluetooth interface, and the device should be reset after that operation.  

It should be better to always explicitly set these details, but I've seen some cases already where it's been very useful to not have to change anything.

PR #875, opened 2018-11-01
change detection for datasheet generation

rickkas7:
PDF generation now keeps a table of the md5 hashes for the source md files (after front-matter and non- pdf-generation sections are removed). This file is committed to Github. When the pdf-generation script is run, it can compare the md5s for the md files to see if the pdf needs to be rebuilt. This not only saves a lot of time, but it eliminates committing pdfs unnecessarily or having to manually discard changes from Github. If you want to rebuild all of the pdfs (after changing the css, for example), just delete scripts/pdf-generation/hashes.json.


monkbroc:
Oh, I'm loving this! Nice thinking Rick!

monkbroc:
Could you add a note to the `README.md` about generating the PDFs and maybe add a command in `package.json` to allow regenerating the PDFs (delete the hash and run the generate command)?

PR #909, opened 2018-12-07
Added clarification when powering the Boron

ninjatill:
Added clarification that the Boron can be powered via the Vusb pin and the acceptable voltage range (copied from the Electron datasheet.)

There will be questions on how to power the Boron from an external power supply until it is clarified. The Gen 2 devices had a Vin pin which better indicated the intent of the pin. See community discussions such as this one: https://community.particle.io/t/powering-electron-via-solar-power/30399/34?u=ninjatill

Verbiage is negotiable but the clarification is still needed IMHO.


pkourany:
Actually, the Vin pin is connected to the Vusb input via a reserve-voltage protection diode.  This diode prevents damage to the powering USB supply when the Boron is also being powered via Vin.  It is the Vin pin that should be used to power the Boron with 3.9V to 12V.

ninjatill:
@pkourany, for mesh devices, and specifically the Boron, the Vin pin is only mentioned in the datasheet block diagram and under the Absolute Maximum Ratings table. The silkscreen on the devices only shows Vusb. Vin is synonymous with Vusb on the diagrams but not in the pin descriptions. Vbus is the USB power connection which is connected to Vusb/Vin through the reverse-voltage protection diode. Why the Vbus at the USB connector and Vusb at the input pin... I have no idea?!?

IMHO, the use of Vusb for the Vin pin was not wise. Many may not look close enough to realize the difference of Vusb vs Vbus. Can't take back the hardware now so I think there needs to be clarification on how to power the device from external power (that is higher than 3.3v).

pkourany:
@ninjatill, Particle will post an official update to the documentation in the very near future.

PR #914, opened 2018-12-12
Fix keep alive typo

daradib:


daradib:
@rickkas7 Thanks for writing this helpful doc by the way

rickkas7:
Thank you for the fix!

PR #921, opened 2019-01-04
Add mention of the publish based ParticleWebLog

geeksville:
Hi, 
I've just written a new LogHandler that allows remote logging using only the standard Particle.publish (which of course has pros and cons).  But it worked well for me because my network is pretty hostile to UDP, I couldn't use ParticlePapertrail.  Instead, I publish to the particle servers and then use a webhook to send my (low rate) log messages to a log server.  During development I just watch the standard particle web console.


geeksville:
cool?  Anything else I need to do to submit this?

PR #922, opened 2019-01-04
The tutorial was a bit stale, 'contribute' is now called 'upload'

geeksville:
The tutorial was a bit stale, 'contribute' is now called 'upload'


rickkas7:
Thank you for the correction!

PR #923, opened 2019-01-09
Topic/event limit still in docs

caljep:
*Don't merge yet*, the webhooks troubleshooting best practice hasn't been refactored --> https://github.com/particle-iot/docs/compare/topic/eventLimitStillInDocs?expand=1#diff-cc717ba54d72579f4ac8848c18dc68b3

@jme783  -- correct me if I'm wrong but the best practice is probably using the `Events` page and monitoring the specific events that get published during webhook integration? Eg `hook-sent/` and `hook-response/`

### DIFF
--- a/src/content/tutorials/device-cloud/console.md
+++ b/src/content/tutorials/device-cloud/console.md
@@ -40,9 +40,9 @@ class="full-width"/>
 You can also take certain actions on devices from this view, such as
 renaming the device and unclaiming it from your account.
 
-### Event Logs
+### Events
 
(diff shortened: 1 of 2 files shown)


jme783:
@caljep — yes correct. You can use the Events page to ensure `hook-sent` and `hook-response` events are coming through properly. It also might be worth mentioning _Integration Audit Logs_ — graphs on the Integration details page that show the performance of the integration over time. 

This feature is described here: https://docs.particle.io/reference/device-cloud/webhooks/#using-the-console

jme783:
Thanks for making this change!

PR #924, opened 2019-01-14
Changed cd firmware to cd device-os

emil01:
The folder created during git clone will be **device-os** not firmware anymore


rickkas7:
Thank you!

PR #929, opened 2019-01-26
Fix battery temperature unit

daradib:


rickkas7:
Thank for for the fix!

PR #934, opened 2019-02-04
Change 7E1 to be even parity

hwestbrook:
Small error in docs describing the even parity option for 7 bit Serial


rickkas7:
Thanks for the fix!

PR #940, opened 2019-02-13
Update Cli.md

bloukingfisher:
reflect that "particle identify" command also gets system firmware version


rickkas7 (approved):
Thank you!

PR #960, opened 2019-03-28
Fix spelling

monkbroc:
I ran `npm run spell` and it caught several misspellings. It's unfortunately quite tedious to do this however since it also catches all of our lingo as misspelling so I tend to do Up arrow -> Enter, Up arrow -> Enter, over and over again to add the lingo words to the dictionary file `.spelling`.


rickkas7 (approved):
Thank you!

monkbroc:
I'll wait for Travis to catch up then merge

PR #972, opened 2019-04-23
Add note about USBSerial1 on Windows for Electron and P1

monkbroc:
Due to limited number of USB endpoints on the Electron and P1 and the unavailability of the lowcdc driver for Windows 10, USBSerial1 can be unreliable on Windows 10.


avtolstoy (approved):
LGTM

PR #988, opened 2019-05-20
Add containerized hosting

zfields:


towynlin:
Regarding that long `docker run` command — it's magical tribal knowledge that's not obvious, therefore (1) thank you so much for putting it in the README :+1:, and (2) please add a script to the `scripts` folder so a human doesn't even need to copy paste from the README.

monkbroc (approved):
Thanks for putting together and demoing the containerized docs Zak!

towynlin:
Thanks! At a glance looks fine — I'll try to run it within the hour.

towynlin:
Running `source scripts/gen-docker-alias.sh` crashes my terminal!

Running `scripts/gen-docker-alias.sh` outputs:
```
\
`getopt --test` failed in this environment.
Please confirm "GNU getopt" is installed on this device.

```

I'm running bash on macOS Mojave 10.14.5 in the normal Terminal app.

zfields:
That's wild! Looks like the script is very Linux centric at this moment.

The error message is accurate. You will need to `brew install gnu-getopt` before the script will ever work, but holy cow that is a failure condition we cannot tolerate.

zfields:
Looks like containerized hosting on Mac may not be an option, until this open Docker for Mac issue gets resolved.
https://github.com/docker/for-mac/issues/2965

towynlin:
If you add language to the readme about the docker for mac issue, we could go ahead and merge this. It's additive, not destructive. Please structure the readme so a user skimming the thing just trying to quickly get started with the docs doesn't go down the wrong path — ~on linux, containerized hosting; on mac, `npm start`; on windows whatever you recommend since I haven't tried it.~ `npm start` should work on all platforms, containerized hosting works on linux not on mac.

zfields:
Awesome! I'll make the update and have you review before merging.

> EDIT: I ended up supplying a note that containerized hosting only works on Linux at this time. Can you have a look and see if you dig it?

PR #990, opened 2019-05-24
Adds simple script to call `docker run` similar to server.js

towynlin:
On my mac, even after letting this script run for 52 minutes, requests to localhost:8080 were still not succeeding. By contrast `npm start` starts serving the docs in about 40 seconds.

### DIFF
--- a/scripts/docker-server.sh
+++ b/scripts/docker-server.sh
@@ -1,12 +1,2 @@
 #!/bin/sh
-
-if [ -z $REPO_PATH ]; then
-    echo "$0 ERROR: The REPO_PATH environment variable is required!"
-    echo "Try this: export REPO_PATH=\`pwd\`"
-    exit 5
-elif ! [ -d $REPO_PATH ]; then
-    echo "$0 ERROR: $REPO_PATH is not a valid directory"
-    exit 6
-fi
-
-docker run --interactive --mount type=bind,source=$REPO_PATH,target=/particle-iot/docs --net host --rm --tty particle/docs
+docker run --interactive --mount type=bind,source=$(pwd),target=/particle-iot/docs --net host --rm --tty particle/docs "$@"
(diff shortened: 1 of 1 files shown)


zfields:
The main reason for the `alias`, is that it allows you pass parameters to the launch script inside the Docker container. Your script calls the container script naked (a safe choice), which is the most thorough and slow version.

Separately, is your Docker VM factory throttled? If you two-finger click on the whale, then you can go into the settings and give it access to more CPU cores and memory. The factory settings for Mac are extremely conservative.

zfields:
I kept digging and I think I have a great mash-up to make it work! I've added it to the review!

scripts/docker-server.sh
```diff
@@ -0,0 +1,12 @@
...
+if [ -z $REPO_PATH ]; then
+    echo "$0 ERROR: The REPO_PATH environment variable is required!"
+    echo "Try this: export REPO_PATH=\`pwd\`"
+    exit 5
+elif ! [ -d $REPO_PATH ]; then
+    echo "$0 ERROR: $REPO_PATH is not a valid directory"
+    exit 6
+fi
+
+docker run --interactive --mount type=bind,source=$REPO_PATH,target=/particle-iot/docs --net host --rm --tty particle/docs
```
zfields:
```bash
#!/bin/bash

# Launch container with forwarded parameters
docker run --interactive --mount type=bind,source=$(pwd),target=/particle-iot/docs --net host --rm --tty particle/docs "$@"
```

zfields (changes requested):

scripts/docker-server.sh
zfields: My proposed solution _ASSUMES_ they are calling the script from the root of the `docs` repository, but would otherwise hide the docker ugliness and allow them to forward the calling options _(without needing to have `gnu-getopt` installed)_.

PR #992, opened 2019-06-19
device-os: clarify description of button_click event

boneskull:
The "setup" button is the same button as the "mode" button.

When I read these docs initially, I asked myself "what's the 'setup' button? I only have 'mode' and 'reset' buttons..."

I figure this might trip someone else up as well.


rickkas7:
Thank you!

PR #995, opened 2019-06-22
workbench tutorial: fix references to "lib"

boneskull:
The tutorial referenced "libs" everywhere, which is incorrect.  

Also normalize formatting in "Project Directories" section, as it was inconsistent.  I can break this PR up if you wish.


rickkas7 (approved):

rickkas7:
Thank you for the fixes!

PR #1013, opened 2019-07-24
feature/cli-updated-dev-info

busticated:


rickkas7 (approved):

rickkas7:
However, your build will likely fail because of a changed link in the store. I'll fix that tomorrow if no one has fixed it first.

busticated:
@rickkas7 ah, ok. no sweat :+1:

i see:

```
The command "./scripts/update-api.sh" failed and exited with 128 during .
```

over in travis - i assume that's the error you are referring to?

rickkas7:
So that's Julien's fix for the changed tag on the api-service. However, that hasn't been released because of a changed store link, which prevent that travis build from succeeding.

busticated:
@rickkas7 got it. i'm ok to merge though, correct?

rickkas7:
Yes, though it won't go live until the previous problems are fixed. I'm about to commit the store fix, which should unblock all of the other changes.

PR #1014, opened 2019-07-30
Document new EventStream behavior

monkbroc:
Story details: https://app.clubhouse.io/particle/story/36413

Adding more details about the behavior of the EventStream class.


monkbroc:
I'll push it out now since particle-api-js 8.0.0 is published.

PR #1016, opened 2019-08-04
updated particle pub docs to use PRIVATE flags in examples

bsatrom:
I've gotten feedback from a few customers that our `Particle.publish` docs can be a bit confusing in that we suggest using `PRIVATE` by default, but the first several examples use the `PUBLIC` flag. This change uses `PRIVATE` for all examples.

### DIFF
--- a/src/content/reference/device-os/firmware.md
+++ b/src/content/reference/device-os/firmware.md
@@ -300,7 +300,7 @@ For the time being there exists no way to access a previously published but TTL-
 
 ---
 
-Publish a public event with the given name, no data, and the default TTL of 60 seconds.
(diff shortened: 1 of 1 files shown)


rickkas7 (commented):
Doesn't this make two examples of PRIVATE and no examples of PUBLIC? There's an example of PRIVATE right after this. Reversing the order, putting PRIVATE first would be fine.

bsatrom:
Yes it does, but if we are recommending PRIVATE by default, why not make them all PRIVATE? Alternatively, we could make the last one private.

bsatrom:
@rickkas7 I changed the last example to use the `PUBLIC` flag. does that work?

PR #1024, opened 2019-09-18
Correct some errors and begin to make return descriptions more consistent

nrobinson2000:


rickkas7:
Looks good, thank you for cleaning that up!

PR #1036, opened 2019-10-23
Tinker updates

bsatrom:
Updated Tinker screenshots and docs for Gen3 and prior devices


rickkas7 (approved):
Looks great, thank you!

PR #1043, opened 2019-11-07
Document new CLI options

monkbroc:


src/content/reference/developer-tools/cli.md
```diff
@@ -960,7 +966,18 @@ New access token expires on Fri Jun 23 2017 23:09:24 GMT+0800 (SGT)
 		da39a3ee5e6b4b0d3255bfef95601890afd80709
 ```
 
-There are no additional options for creating a token from the CLI. In order to change the expiration or make a non-expiring token, you should use the Particle Cloud API directly.
+In order to change the time a token is valid, use `--expires-in <seconds>` option.
```
wraithan:
```suggestion
In order to change the duration a token is valid, use `--expires-in <seconds>` option.
```

I think duration is a better fit.

wraithan (approved):
Have a minor suggestion, but overall looks good, thanks for documenting those additions!

monkbroc:
That's all there is to docs. `npm install` and `npm start` to see how it looks.

PR #1055, opened 2019-12-17
feature/workbench-macos-catalina-alert

busticated:


rickkas7 (approved):
Could you change "macOS Catalina" to "Mac OS 10.15 (Catalina)" so it matches the style of the line right above it? Or at least "macOS" to "Mac OS".

PR #1065, opened 2020-01-27
Update api service docs to src instead of lib

wraithan:
This got missed when `api-service` got reorganized.


monkbroc (approved):
:+1: go ahead and merge it Wraithan

PR #1067, opened 2020-01-28
Document adding custom symbols during Device OS application build

monkbroc:
Story details: https://app.clubhouse.io/particle/story/45807

I'll be distributing this nugget of info after it's merged to docs.particle.io


rickkas7 (approved):
Thanks! That's clever, I never thought about doing that before.

PR #1072, opened 2020-01-30
Fix internal links to gen3 accessories

monkbroc:
Redirect doesn't maintain the anchor so links to accessories went to the top of the page. Change the internal links to point to the new page

Rick, can you please merge when Travis passes?


rickkas7 (approved):
Thanks! Will merge when done.

PR #1077, opened 2020-02-07
Remove documentation about Firmware Manager

monkbroc:
Firmware Manager is completely out of date and not useful anymore so remove it from the docs to avoid confusing people.


rickkas7 (approved):
Thank you!

PR #1087, opened 2020-03-05
Update workbench.md

dhhagan:
After completing the first three steps, you should already have the 'develop' branch so do not need to create it


rickkas7 (approved):

rickkas7:
Thank you!

PR #1098, opened 2020-04-30
Update the library link for the Grove Ultrasonic Ranger

jaredpsimpson:
I noticed that the library link for the Grove Ultrasonic Ranger for the Web IDE kept spinning, so I found the location of the most recent library and updated the link.


rickkas7 (approved):
Thank you!

PR #1106, opened 2020-06-29
Tracker SoM documentation

rickkas7:
Overview of new docs can be found here:
https://docs.particle.io/tutorials/asset-tracking/introduction/#find-out-more


kennethlimcp:
@rickkas7 for some reason the deployment doesnt seem to upload this image which is messing in the docs:

https://docs.particle.io/assets/images/at-som/M8-cable.jpg

rickkas7:
The fix should be live in a few minutes. It's because when deployed to production it runs on Linux so the file system is case-sensitive and the file is m8-cable.jpg. It works and passes the link vertification on the Mac because the file system is not case-sensitive.

PR #1119, opened 2020-09-24
Update aqmk-project.md - Fixes issues with compile errors and an incorrect #include

Landsailor:
Corrects a few issues that I encountered while assembling my AQM kit:

1. Lack of #include <math.h> (Perhaps that's handled by the #include "Particle.h" in the final documentation. If so, then it should be included there instead.) 
2. A duplicate of the #include "Air_Quality_Sensor.h" for the Adafruit_BME280 section.


rickkas7:
Thank you!

PR #1123, opened 2020-10-14
Update note in write() section of TCPClient

monkbroc:
Story details: https://app.clubhouse.io/particle/story/62086

Ensure timeout parameter for TCPServer.write is shown for all platforms and remove confusing statement about write returning negative error codes now that 0.9.0 has been out for a while.


rickkas7 (approved):
Thank you!

PR #1129, opened 2020-11-13
Scrub SPI docs from using SPI.transfer without transaction

monkbroc:
Story details: https://app.clubhouse.io/particle/story/52132


rickkas7 (approved):
Thank you!

PR #1133, opened 2020-12-02
Add 2.0.0 Device OS version to docs

avtolstoy:
Also replaces 2.0.0-rc.4 references with 2.0.0 in https://docs.particle.io/reference/device-os/versions/


rickkas7 (approved):
Thank you!

PR #1135, opened 2020-12-10
Add config proxyUrl example

yrtimiD:
Add proxyUrl setting example to the `particle config` command usage


yrtimiD:
Guys, do you accept PR's?

PR #1140, opened 2021-01-26
Fixes to CLI ref

ajsirota:
- Better explain `nyan`
    - Change examples from `particle cloud nyan` to just `particle nyan`
- Fix grammar in `token create`


rickkas7 (approved):

rickkas7:
Thank you!

PR #1169, opened 2021-04-23
Minor changes in the  RX part of the cellular.

hannanmustajab:
In the documentation for getDataUsage, I think it should be sent and received instead of both being sent.


rickkas7:
Thank you!

PR #1175, opened 2021-04-28
Fix 'enhanced_loc' -> 'loc-enhanced', other style improvements

scottbez1:
From testing, it appears the enhanced location event is actually `loc-enhanced` not `enhanced_loc` as documented here. Fixed that and improved style of other event keys to format as code. Also fixed `loc.bat` which I believe should actually be `loc.batt` based on tracker-edge v13.


rickkas7:
Thank you!

PR #1176, opened 2021-04-28
Fix "loc.src" -> "src" in enhanced location docs

scottbez1:
From testing, it appears that the `src` field is added at the top level, not nested within the `loc` object


rickkas7:
I haven't merged this because there is a possibility that the docs are right and the implementation is wrong.

PR #1200, opened 2021-06-30
Fix broken Icomoon font link

monkbroc:
Fixes a 404 not found error on the docs trying to load the Icomoon font. This is an icon font used in the Console. The S3 link is no longer available and it must be loaded from Cloudfront.

Steps to test:
- Check that https://docs.particle.io/tutorials/product-tools/device-groups/#creating-device-groups shows the group icon in the parentheses.
image


rickkas7 (approved):
Thank you!

PR #1216, opened 2021-09-14
Added missing ; in getSoC example

trylaarsdam:
Added the missing semicolon to the getSoC prototype in the example:

original:
```cpp
float getSoC() 
```
modified:
```cpp
float getSoC();
```


rickkas7:
Thank you!

PR #1219, opened 2021-09-16
remove redundant include

rcaudill:


rickkas7:
I see where it's confusing, but it's not actually redundant. The examples list the include header that's required to use the class, followed by a full example, which includes everything you need to compile the sample. It would be less confusing as two separate text boxes, but in the three column layout that's not really possible, unfortunately.

PR #1237, opened 2022-01-21
Document particle serial wifi --file

monkbroc:
The automation for `particle serial wifi --file credentials.json` wasn't documented so I added it.


rickkas7 (approved):
Thank you!

PR #1265, opened 2022-05-18
Remove Desktop IDE (Dev) Page

monkbroc:
Story details: https://app.shortcut.com/particle/story/43423

At this point, we ought to just remove the Particle Dev page from the docs and redirect to Workbench.

Steps to validate:
- https://docs.particle.io/reference/discontinued/software/dev/ redirects to https://docs.particle.io/getting-started/developer-tools/workbench/


rickkas7 (approved):
Makes sense - thank you!

PR #1266, opened 2022-05-18
Improve contrast on api docs page

monkbroc:
The request/response sections and types in the API docs were hard to read in pale gray so I made them darker.

Before:
image

After:
image


rickkas7 (approved):
Thank you!

PR #1294, opened 2022-08-30
add Device OS 4.0.0 (LTS)

avtolstoy:
@rickkas7 I haven't found scripts that help make zips/json for device restore. I'll leave this as-is for now, but might be a good idea to add a readme on that, so that we can create that as part of the release without having to bug you :smile_cat:


technobly (approved):

rickkas7 (approved):
There are instructions in Notion but there are manual steps. It's on my list to work with Mirande to make it an automated step in the build so you all don't have to do yet another manual step during release.

PR #1328, opened 2023-01-12
feat: [sc-114387] Fix API docs capitalization

monkbroc:
Story details: https://app.shortcut.com/particle/story/114387


monkbroc:
Asking for a review since it's my first update to shared blurbs. The config needs to be check in, right?

rickkas7 (approved):
Thanks! Looks good. I'll write a script to process the rest of the markdown headers in the docs.

PR #1366, opened 2023-06-22
Add PRODUCT_ID variable to webhook docs

wraithan:
Took the liberty of doing the docs update myself, no rush to review/get this out until you're back, just figured I could take a stab.


rickkas7 (approved):
Looks good. I'll merge it now and it will be live in ~10 minutes

PR #1370, opened 2023-07-20
Update to Asset OTA API

monkbroc:
Here are a few tweaks to the Asset OTA docs. I added a link to the repo with examples. Feel free to move the link to another "Asset OTA tutorial" page when that is ready.


rickkas7 (approved):
Thank you!

PR #1375, opened 2023-07-27
Add Photon 2 pinout to You're muted example

ishotjr:


rickkas7:
The change makes sense. Thanks!

I manually applied the change instead of merging since I needed to change the path to be relative.

PR #1391, opened 2023-09-25
Bugfix: undefined gtag

davidbbe:
Segment is no longer loading the gtag function in the browser even if users allow cookies. This was my mistake since Segment says on the docs "[Segment loads the gtag.js library for you](https://segment.com/docs/connections/destinations/catalog/actions-google-analytics-4-web/)". We need to change all of the gtag events to the Segment version that does the same thing. 

This needs to change:
`gtag('event', 'Event name', {'event_category':eventCategory});`
To this Segment event:
`analytics.track('Event name', {category:eventCategory});`

The `analytics` object is always available in the browser even if the user denies cookies. So if the user denies cookies and a `analytics.track` tries to happen, there wont be any errors.

Here is the migration guide for setting up any future events with Google Analytics https://segment.com/docs/guides/how-to-guides/migrate-from-other-tools/#custom-events


rickkas7 (approved):
Looks good and appears to work locally for me

PR #1425, opened 2024-02-15
Device-side ledger documentation

rickkas7:


src/content/getting-started/logic-ledger/ledger.md
```diff
@@ -93,7 +103,75 @@ In the **Advanced** tab, you can set arbitrary JSON data, including nested objec
...
 {{imageOverlay src="/assets/images/ledger/ledger-advanced.png" class="no-darken"}}
 
-When using the Cloud API you get back the entire JSON data structure for a ledger instance.
+When using the [Cloud API](/reference/cloud-apis/api/#ledger) you get back the entire JSON data structure for a ledger instance.
+
+## Cloud ledger - console
+
+- Stores information about a device, product or your entire fleet in the cloud.
+- Data can be read and written by your applications through the Particle API and Logic Functions.
+- Cloud Ledgers cannot be read by a device.
```
sergeuz, line 105: How about we add a matrix to summarize it in one place who can read or write what for different ledger types in addition to the descriptions here?

src/content/reference/device-os/firmware.md
```diff
@@ -781,6 +781,32 @@ Particle.unsubscribe();
...
+
+{{api name1="Particle.ledger"}}
+
+{{since when="5.8.0"}}
+
+Request a ledger from the cloud. Returns a [`Ledger`](#ledger) object to read or write the data in the ledger. 
+
+This operation is asynchronous and the data will not be available until synchronized with the cloud. You will typically call this from `setup()` and you can do so before connected to the cloud.
+
+The type of ledger (device-to-cloud or cloud-to-device), as well as the ledger scope (organization, product, or device) is determined when the ledger is created on the cloud side, so it is not specified when you request the ledger. You must first create a ledger in the cloud; you cannot create a new ledger using the device-side API. The first time a device comes online in a product with a device-specific ledger, a new ledger instance will be created automatically for the device, however.
```
sergeuz: > You must first create a ledger in the cloud; you cannot create a new ledger using the device-side API.

> The first time a device comes online in a product with a device-specific ledger, a new ledger instance will be created automatically for the device, however.

This paragraph is confusing because the docs don't make clear distinction between ledger _definitions_ and _instances_.

src/content/reference/device-os/firmware.md
```diff
@@ -12979,6 +13005,1754 @@ int8_t measurePower() const;
...
+
+You will typically store the `Ledger` object you receive from `Particle.ledger()` as a global variable so you can access it again later easily.
+
+Ledger names consist only of lowercase alphanumeric and dash, up to 32 characters, and are unique across all scopes.
+
+
+
+### set() [Ledger class]
+
+Sets a value in the ledger. This call is asychronous.
```
sergeuz: It's worth documenting that, locally, the data gets saved immediately. It's the synchronization with the cloud that happens asynchronously and, potentially, at a later time.

src/content/reference/device-os/firmware.md
```diff
@@ -12979,6 +13005,1754 @@ int8_t measurePower() const;
...
+```cpp
+// PROTOTYPE
+static int remove(const char* name);
+```
+
+### removeAll() [Ledger class]
+
+Remove any ledger data from the device.
+
+The device must not be connected to the Cloud. The operation will fail if the ledger with the given name is in use.
```
sergeuz:
```suggestion
The device must not be connected to the Cloud. The operation will fail if any of the ledgers are in use.
```

src/content/reference/device-os/firmware.md
```diff
@@ -12979,6 +13005,1754 @@ int8_t measurePower() const;
...
+void loop() {
+}
+
+```
+
+## LedgerData
+
+The `LedgerData` class provides a subset of methods of the `Variant` class that are relevant to map operations. The Ledger class [set()](#set-ledger-class-) and [get()](#get-ledger-class-) take a LedgerData object. 
+
+Because a `Variant` is transparently converted into `LedgerData` you may never need to create a `LedgerData` object. See [Ledger sensor](/getting-started/logic-ledger/ledger-sensor/) for an example of this technique.
```
sergeuz, line 13389: This is not true for a conversion in the opposite direction though. I'm wondering if we should allow `LedgerData` to be transparently converted to a `Variant` as well.

src/content/reference/device-os/firmware.md
```diff
@@ -12979,6 +13005,1754 @@ int8_t measurePower() const;
...
+data.set("sensor", sensorValue);
+data.set("time", Time.format(TIME_FORMAT_ISO8601_FULL));
+sensors.set(data);
+```
+
+See [Ledger sensor](/getting-started/logic-ledger/ledger-sensor/) for the full example.
+
+### Variant::Type
+
+Variants have an explicit type, unlike JSON. The following
```
sergeuz, line 13524:
```suggestion
Variants have an explicit type, unlike JSON. The following types are supported:
```

src/content/reference/device-os/firmware.md
```diff
@@ -12979,6 +13005,1754 @@ int8_t measurePower() const;
...
+Variant(String val);
+Variant(VariantArray val);
+Variant(VariantMap val);'
+```
+
+### value() [Variant class]
+
+{{api name1="Variant::value()"}}
+
+The `value` method is a template that returns a reference to the stored value of the variant. The allowable types are listed in [Variant::Type](#variant-type). This can be done to either get or set a value.
```
sergeuz, line 13570: It should be documented that the template type of this method must match the actual underlying type of the variant value, otherwise the behavior is undefined. A safer alternative is to use the `toXXX()` and `asXXX()` methods.

src/content/reference/device-os/firmware.md
```diff
@@ -12979,6 +13005,1754 @@ int8_t measurePower() const;
...
+bool isArray() const;
+bool isMap() const;
+```
+
+
+### toBool() [Variant class]
+
+{{api name1="Variant::toBool()"}}
+
+Returns value of the variant to an `bool` (boolean, `true` or `false`), converting the type if necessary. The original value is left unchanged.
```
sergeuz, line 13639:
```suggestion
Returns value of the variant as a `bool` (boolean, `true` or `false`), converting the type if necessary. The original value is left unchanged.
```

src/content/reference/device-os/firmware.md
```diff
@@ -12979,6 +13005,1754 @@ int8_t measurePower() const;
...
+
+| Source | Result | 
+| :--- | :--- |
+| bool | unchanged |
+| numeric | zero (`0`) &rarr; `false`, non-zero &rarr; `true` |
+| String | `false` &rarr; `false`, 'true' &rarr; "true" |
```
sergeuz, line 13645:
```suggestion
| String | "false" &rarr; `false`, "true" &rarr; `true` |
```

src/content/reference/device-os/firmware.md
```diff
@@ -12979,6 +13005,1754 @@ int8_t measurePower() const;
...
+{{api name1="Variant::asBool()"}}
+
+Returns a reference to the value contained in this variant as a `bool`. This can be used to modify the value of the variant.
+
+```cpp
+// PROTOTYPE
+bool& asBool();
+```
+
+This method will convert the type of the variant to bool if necessary, see [toBool](#tobool-variant-class-) for more information. If the conversion is not possible, the method will assert.
```
sergeuz: It's not clear what's the difference between `value<bool>()`, `asBool()` and `toBool()`. Perhaps it's worth adding a section about that so that we don't have to document it for each of these methods separately.

- `value<T>()` – returns a reference to the currently held value. Requires it be of the type `T` otherwise would cause undefined behavior.
- `as<T>()` or `asXXX()` – converts the underlying value of the variant to `T` _in place_, modifying the variant, and returns a reference to it. Safe to call if the types don't match but requires `T` to be one of the supported types, otherwise would cause a compilation error (not a runtime assert as currently documented).
- `to<T>()` or `toXXX()` – converts the underlying value to `T` without modifying the variant and returns the result _by value_. Safe to call if the types don't match, `T` can be arbitrary.

src/content/reference/device-os/firmware.md
```diff
@@ -12979,6 +13005,1754 @@ int8_t measurePower() const;
...
+
+### toString() [Variant class]
+
+{{api name1="Variant::toString()"}}
+
+Returns the value of the variant to an `String` (an ASCII or UTF-8 string) and returns a copy of it, converting the type if necessary. The original value is left unchanged.
+
+| Source | Result | 
+| :--- | :--- |
+| bool | false &rarr; "0", true &rarr; "1" |
```
sergeuz, line 13888:
```suggestion
| bool | false &rarr; "false", true &rarr; "true" |
```

src/content/reference/device-os/firmware.md
```diff
@@ -12979,6 +13005,1754 @@ int8_t measurePower() const;
...
+// EXAMPLE
+const char *jsonStr = "{\"a\":123}";
+Variant data = Variant::fromJson(jsonStr);
+```
+
+## VariantArray
+
+{{api name1="VariantArray"}}
+
+The `VariantArray` is a [Vector](#vector), essentially a dynamically-sized array of Variants, which are containers for arbitrary data. Think of this as a container to hold arbitrary data in Variant objects before converting to a JSON or CBOR array before sending across the network.
```
sergeuz: > or CBOR array before sending across the network

I'd probably remove all mentions of CBOR from the docs for simplicity.

src/content/reference/device-os/firmware.md
```diff
@@ -12979,6 +13005,1754 @@ int8_t measurePower() const;
...
+
+```
+// DEFINITION
+typedef Vector<Variant> VariantArray;
+```
+
+The methods below are implemented in the `Variant` class, so you can use them update a `Variant` that is of type `VariantArray`.
```
sergeuz: > you can use them update a `Variant` that is of type `VariantArray`

The array methods defined in `Variant` will convert the variant to an array so it's not required for the variant to be an array prior to the operation. The map methods will convert the variant to a map.

src/content/reference/device-os/firmware.md
```diff
@@ -12979,6 +13005,1754 @@ int8_t measurePower() const;
...
+```
+
+### get() [VariantMap]
+
+{{api name1="Variant::get(key)"}}
+
+Get the value of an element in the map. This method is implemented in `Variant`.
+
+This method copies the value, which is inefficient for complex variant values. operator[] can be used
+to get a reference to the value instead of copying it.
```
sergeuz, line 14137: Let's add a similar note for `at()` that returns the element of the array by value.

src/content/getting-started/logic-ledger/ledger.md
rickkas7: Added a table

src/content/reference/device-os/firmware.md
rickkas7: Changed to:

The type of ledger (device-to-cloud or cloud-to-device), as well as the ledger scope (organization, product, or device) is determined when the ledger is created on the cloud side, so it is not specified when you request the ledger. You must first create a ledger definition in the cloud; you cannot create a new ledger definition using the device-side API. 

The first time a device comes online specifying a device to cloud ledger, a new ledger instance will be created for the device, however.

src/content/reference/device-os/firmware.md
rickkas7: Changed to: The data is saved to local flash storage immediately, but will be synchronized to the cloud at a later time.

src/content/reference/device-os/firmware.md
rickkas7: Updated

src/content/reference/device-os/firmware.md
rickkas7: Added a new section with this explanation, and linked to it from all of the methods.

src/content/reference/device-os/firmware.md
rickkas7: Fixed

src/content/reference/device-os/firmware.md
rickkas7: Done

src/content/reference/device-os/firmware.md
rickkas7: Fixed

src/content/reference/device-os/firmware.md
rickkas7: Fixed

PR #1447, opened 2024-05-16
feat: [sc-127947] Update API docs to include query and body parameters

monkbroc:
Story details: https://app.shortcut.com/particle/story/127947

Updates the format of the API docs generation to separate URL parameters like :deviceId from query params like :page and request attributes like :notes.

Replaces all examples of access token in the query string with access token in the Authentication header since this is more secure. Access token in the query string is marked as deprecated and discouraged.

Also fixes docs generation for particle-api-js that was due to introducing Typescript annotations.

Steps to test:
- [X] Verify that all the existing API endpoints still show up in the documentation

### DIFF
--- a/config/sitemap.json
+++ b/config/sitemap.json
@@ -1155,8 +1155,8 @@
             "hash": "74b9315484be16ea70298a49d15ff440f68b636eb3074af419bb725240c7204e"
         },
         "reference/cloud-apis/javascript.md": {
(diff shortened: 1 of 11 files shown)


rickkas7 (approved):
Looks good! I mostly looked at the non-content changes, but also scanned the content diffs and ran it and scrolled through the whole thing and it looks good to me. Thank you!

PR #1452, opened 2024-06-07
Update device-constants for the updated platform display names

keeramis:
Display names and generation numbers are updated in device-constants 3.4.0

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,6 +9,7 @@
             "version": "0.3.0",
             "license": "CC-BY-SA-3.0",
             "dependencies": {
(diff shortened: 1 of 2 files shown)


rickkas7 (commented):
Docs doesn't consume device-constants directly; it does it through CarrierListTool which generates a json file with various device-specific data, one of which is device constants. I updated CarrierListTool and docs with the updated version so this PR isn't needed.

keeramis:
Awesome, thanks @rickkas7 . This is just for the display names and generations. I went through git and found that docs was using the displaynames from the device constants

PR #1474, opened 2024-10-11
Feature/typed publish

rickkas7:


src/content/reference/device-os/firmware.md
```diff
@@ -405,1061 +405,1306 @@ curl https://api.particle.io/v1/devices/0123456789abcdef/brew \
...
 
-A `system_error_t` result code
+{{!-- 
+Additionally, JSON (text/json) structured data is supported. This is encoded using an `EventData` object, 
+which is a `Variant`. The `Variant` can be created from a JSON text string, or built programmatically
+by adding fields.
 
-* `system_error_t::SYSTEM_ERROR_NONE`
-* `system_error_t::SYSTEM_ERROR_IO`
+If you are currently publishing events using a JSON string, it's highly recommended that you switch
```
sergeuz, line 1160: > it's highly recommended that you switch to using `Variant`

There's a caveat that `Variant` may be less memory efficient for relatively large payloads than, for example, `JSONBufferWriter` or string interpolation. We should perhaps mention that and let users decide what approach fits their use case better.

src/content/reference/device-os/firmware.md
```diff
@@ -405,1061 +405,1306 @@ curl https://api.particle.io/v1/devices/0123456789abcdef/brew \
...
 
-  if ( condition ) {
-    Particle.publishVitals();  // Publish vitals immmediately
-  }
-}
+```cpp
+// PROTOTYPES
+particle::Future<bool> publish(const char* name, const char* data, particle::ContentType type, PublishFlags flags = PublishFlags());
+particle::Future<bool> publish(const char* name, const String& data, particle::ContentType type, PublishFlags flags = PublishFlags());
+particle::Future<bool> publish(const char* name, const char* data, size_t size, particle::ContentType type, PublishFlags flags = PublishFlags());particle::Future<bool> publish(const char* name, const particle::EventData& data, PublishFlags flags = PublishFlags());
```
sergeuz, line 1184:
```suggestion
particle::Future<bool> publish(const char* name, const char* data, size_t size, particle::ContentType type, PublishFlags flags = PublishFlags());
particle::Future<bool> publish(const char* name, const particle::EventData& data, PublishFlags flags = PublishFlags());
```

src/content/reference/device-os/firmware.md
```diff
@@ -405,1061 +405,1306 @@ curl https://api.particle.io/v1/devices/0123456789abcdef/brew \
...
-loop () {
-  // Cancel vitals after one hour
-  if (3600000 < (millis() - start)) {
-    Particle.publishVitals(0);  // Publish immediately and cancel periodic publishing
-  }
-}
-```
+In Device OS 6.2 and later, the data can be specified with a pointer and length. This is typically used for binary
+data, and in particular, data that can contain a null byte. It can be used for text if your data source is not
+null-terminated.
```
sergeuz, line 1202: :+1:

src/content/reference/device-os/firmware.md
```diff
@@ -405,1061 +405,1306 @@ curl https://api.particle.io/v1/devices/0123456789abcdef/brew \
...
 
-{{since when="1.5.0"}}
+#### data (EventData) - Particle.publish - Publish
 
-You can also specify a value using [chrono literals](#chrono-literals), for example: `Particle.publishVitals(1h)` for 1 hour.
+The `EventData` overload does not `ContentType` value because it is used for structured data in a `Variant` and is always
```
sergeuz, line 1206:
```suggestion
The `EventData` overload does not require a `ContentType` value because it is used for structured data in a `Variant` and is always
```

src/content/reference/device-os/firmware.md
```diff
@@ -405,1061 +405,1306 @@ curl https://api.particle.io/v1/devices/0123456789abcdef/brew \
...
+// EXAMPLE 3
+Future<bool> futureResult = Particle.publish("myEvent");
+```
 
-{{api name1="Particle.connect"}}
+In Example 1, you do not care about whether the call succeeds or not, and this will not wait for the data to be acknowledged by the cloud.
 
-`Particle.connect()` connects the device to the Cloud. This will automatically activate the network connection and attempt to connect to the Particle cloud if the device is not already connected to the cloud.
+In Example 2, you can check bResult. If true, the publish was successfully made to the cloud. This will block until the packet is acknowledged
+by the cloud.
```
sergeuz, line 1241: > If true, the publish was successfully made to the cloud

This is not guaranteed unless `publish()` was called with a `WITH_ACK` flag.

src/content/reference/device-os/firmware.md
```diff
@@ -405,1061 +405,1306 @@ curl https://api.particle.io/v1/devices/0123456789abcdef/brew \
...
-It will also update the [ApplicationWatchdog](/reference/device-os/api/watchdog-application/watchdog-application/) timer using `ApplicationWatchdog::checkin()`.
 
-### Particle.syncTime()
+### Subscribe (with content type) - Subscribe
 
-{{api name1="Particle.syncTime"}}
+```cpp
+// PROTOTYPES
+bool subscribe(const char* name, particle::EventHandlerWithContentType handler);
+bool subscribe(const char* name, particle::EventHandlerWithContentTypeFn handler);
```
sergeuz, line 1496: > particle::EventHandlerWithContentType
> particle::EventHandlerWithContentTypeFn

Nit: Can we not expose these type names and just document it by example what callback signatures are expected? Definitely feel free to keep it the way it is if you feel doing so would complicate things anyhow. It's just that I would like to keep some room for renaming and moving things around as this API evolves.

src/content/reference/device-os/firmware.md
```diff
@@ -405,1061 +405,1306 @@ curl https://api.particle.io/v1/devices/0123456789abcdef/brew \
...
+The new API passes both a `data` and a `size` allowing it to be used with data that contains null bytes,
+including binary data. The `ContentType` is also passed to the handler for the handler to use as desired.
 
-Synchronizing time does not consume Data Operations from your monthly or yearly quota. However, for cellular devices they do use cellular data, so unnecessary time synchronization can lead to increased data usage, which could result in hitting the monthly data limit for your account.
 
-For more information about real-time clocks on Particle devices, see [Learn more about real-time clocks](/reference/device-os/real-time-clocks/).
+### Subscribe (with Variant) - Subscribe
 
-### Particle.syncTimeDone()
+If you are expecting structured data, use the subscribe overloada that takes a handler with a Variant.
```
sergeuz, line 1514:
```suggestion
If you are expecting structured data, use the subscribe overload that takes a handler with a Variant.
```

sergeuz (approved):
Looks good :+1: The diff is difficult to navigate due to various sections being moved around so I mostly reviewed the newly added "Publish" and "Subscribe" sections as well as the typed publish page.

src/content/reference/device-os/firmware.md
rickkas7: Removed that sentence.

src/content/reference/device-os/firmware.md
rickkas7: Fixed

src/content/reference/device-os/firmware.md
rickkas7: Updated the example to use WITH_ACK

src/content/reference/device-os/firmware.md
rickkas7: I put an example first. The full prototypes are listed farther down now.

src/content/reference/device-os/typed-publish.md
```diff
@@ -0,0 +1,177 @@
...
+does not use additional data over-the-air, as was the case when encoding to hex, Base64, or Base85 on-device. When the cloud 
+receives the binary data, it expands it into a JSON object containing meta information and Base64 encoded data. This allows
+for easy processing in Logic, webhooks, or SSE.
+
+For example, you could use this code:
+
+```cpp
+VariantMap map;
+map.set("a", Variant(1234));
+map.set("b", Variant(Buffer::fromHex("9dcae4dfe9af57338b5755d67e51771c")));
```
sergeuz, line 118: This example can be simplified:

```cpp
Variant map;
map.set("a", 1234);
map.set("b", Buffer::fromHex("9dcae4dfe9af57338b5755d67e51771c"));
```
For convenience, `Variant` exposes some of the methods of `VariantMap` and `VariantArray` so that it's not necessary to cast to a different type when working with maps and arrays. Also, all constructors of `Variant` are implicit so there's usually no need to call them explicitly.

rickkas7: Oh, that's really useful! Fixed.

src/content/reference/device-os/firmware.md
```diff
@@ -13782,6 +14066,220 @@ bool isEmpty() const;
...
+
+{{api name1="Buffer"}}
+
+{{since when="6.2.0"}}
+
+The `Buffer` class is a container for data in a dynamically allocated buffer. It's essentially a wrapper
+around `std::vector<char>` with additional convenience methods for encoding the data.
+
+The `Variant` class has been extended to store arbitrary binary data in a `Buffer`. A Variant containing a `Buffer` 
+cannot be serialized to JSON or deserialized from JSON as JSON does not support binary values.
```
monkbroc, line 14079: Is this statement true? Won't a Variant containing a Buffer be serialized with `{ "_type": "Buffer", "_data": "11aabbee22" }`

src/content/tools/developer-tools/json.md
```diff
@@ -29,4 +29,19 @@ The five buttons only work when the JSON is valid. If there are syntax errors, t
...
 tool below will convert the JSON above into code.
 
-{{> json-to-code rows="15" cols="90"}}
+{{> json-to-code rows="10" cols="90"}}
+
+### Convert to CBOR
+
+CBOR is a compact binary representation of structured data that can be converted to and from JSON.
+
+{{> json-to-cbor rows="10" cols="90"}}
```
monkbroc, line 38: What's the use case for this tool? We don't expose CBOR directly to the user. It's a transport optimization for Variant.

monkbroc (approved):
Looks great! I'll let you resolve the final comments with Sergey.

src/content/tools/developer-tools/json.md
rickkas7: It's so users can see how big they can make their data and still have it fit in the current 1024 byte limit.

src/content/reference/device-os/firmware.md
rickkas7: I thought that the Variant with buffer to JSON occurred on the cloud-side, not when using the toJSON() method of Variant, but I could have misunderstood. However, my wording doesn't make it clear that's what I meant, so I'll fix that regardless.

sergeuz: > I thought that the Variant with buffer to JSON occurred on the cloud-side, not when using the toJSON() method of Variant

That's correct. `toJSON()` doesn't use the same conversion rules as the cloud does when a variant is published which may indeed cause some confusion. I'd probably just remove this paragraph about `Variant` given that the docs for `toJSON()` already mention that it can't serialize values of certain types.

I thought about using the same conversion rules both in Device OS and the cloud but wasn't keen on pulling Base64 code in Device OS.

PR #1488, opened 2025-01-28
Extended publish docs 6.3.0

rickkas7:


src/content/reference/device-os/firmware.md
```diff
@@ -1439,6 +1466,444 @@ Previously, there were overloads with a `ttl` (time-to-live) value. These have b
...
+| `ContentType::TEXT`   | `text/plain; charset=utf-8` | 0 |
+| `ContentType::JPEG`   | `image/jpeg` | 22 |
+| `ContentType::PNG`    | `image/png` | 23 |
+| `ContentType::BINARY` | `application/octet-stream` | 42 |
+| `ContentType::STRUCTURED` | | 65001 |
+{{!-- END shared-blurb --}}
+
+When publishing you typically set the value, but you can also get the value that was previously set.
+
+When subscribing to an event, you typically get the value to see what the publisher or the event has set.
```
sergeuz, line 1520:
```suggestion
When subscribing to an event, you typically get the value to see what the publisher of the event has set.
```

src/content/reference/device-os/firmware.md
```diff
@@ -1439,6 +1466,444 @@ Previously, there were overloads with a `ttl` (time-to-live) value. These have b
...
+
+You don't typically sent the content type to `STRUCTURED`; when you set the data from a `Variant` this is handled automatically for you.
```
sergeuz, line 1522:
```suggestion
You don't typically set the content type to `STRUCTURED`; when you set the data from a `Variant` this is handled automatically for you.
```

src/content/reference/device-os/firmware.md
```diff
@@ -1439,6 +1466,444 @@ Previously, there were overloads with a `ttl` (time-to-live) value. These have b
...
+Copies data from a `String` object.
+
+```cpp
+// PROTOTYPE
+CloudEvent& data(const String& data)
+```
+
+#### dataString - CloudEvent
+
+Returns a copy of the data in the `EventData` object as a `String`. This is typically used for text data.
```
sergeuz, line 1571:
```suggestion
Returns a copy of the data as a `String`. This is typically used for text data.
```

src/content/reference/device-os/firmware.md
```diff
@@ -1439,6 +1466,444 @@ Previously, there were overloads with a `ttl` (time-to-live) value. These have b
...
+Set the size of the data. This is rarely needed.
+
+```cpp
+// PROTOTYPE
+int setSize(size_t size)
+```
+
+#### size - CloudEvent
+
+Get the size of the data. This is the encoded size so there is computation involved in getting this value.
```
sergeuz, line 1655: Under the hood, the data is stored in an encoded form so the size is always known and doesn't require computation when queried via the underlying CoAP API.

src/content/reference/device-os/firmware.md
```diff
@@ -1439,6 +1466,444 @@ Previously, there were overloads with a `ttl` (time-to-live) value. These have b
...
+based on the current rate limiting.
+
+```cpp
+// PROTOTYPE
+size_t size() const
+```
+
+#### isEmpty - CloudEvent
+
+Returns true if the event is empty  (no has data).
```
sergeuz, line 1667:
```suggestion
Returns true if the event is empty (has no data).
```

sergeuz (approved):
Looks great! I suggested a few fixes, mostly for typos.

PR #1548, opened 2026-02-24
Environment variables

rickkas7:


src/content/getting-started/configuration/environment.md
```diff
@@ -0,0 +1,306 @@
...
+
+| Band | Bit  | Mask |
+| :--- | :--- | :--- |
+| B1   | 0    | 1    |
+| B2   | 1    | 2    |
+| B3   | 2    | 4    |
+| B4   | 3    | 8    |
+| B5   | 4    | 16   |
+
+The value to store for the key is a uint128 value represented in decimal of the band mask values added together.
```
technobly: "represented in hexidecimal"

Would be nice to show it's safe to force enable all bands with "FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF" as they will be further masked against the defaults for each modem.

src/content/getting-started/configuration/environment.md
```diff
@@ -0,0 +1,306 @@
...
+{{> band-mask-calculator }}
+
+
+### PARTICLE_CELLULAR_FORBIDDEN_BANDS
+
+- Available in Device OS 6.4.0 and later
+- Gen 3 and Gen 4 devices
+- Cellular Modems: BG95-M5, BG96-MC, EG91-E, EG91-EX, EG91-NA, R410, R510, SG560D
+
+Sets a mask of bands to not use. The band mask is the same uint128 format as for `PARTICLE_CELLULAR_PREFERRED_BANDS`.
```
technobly: Would be nice to show it's safe to force disable all forbidden bands with "0".  Trying to disable all bands will cause an error on the device.

src/content/getting-started/configuration/environment.md
```diff
@@ -0,0 +1,306 @@
...
+
+{{box op="start" cssClass="boxed warningBox"}}
+Changing cellular environment variables may adversely affect connectivity. Under normal circumstances you should never need to change these values, and if you do change them, be sure to test changes on a subset of your fleet that you have easy access to in case the changes make the device unable to connect to cellular again.
+{{box op="end"}}
+
+### PARTICLE_CELLULAR_PREFERRED_PLMN
+
+- Available in Device OS 6.4.0 and later
+- Gen 3 and Gen 4 devices
+- Cellular Modems: BG95-M5, BG96-MC, EG91-E, EG91-EX, EG91-NA, R410, R510, SG560D
```
technobly: Please remove R410 and BG96-MC from this list.  I found out it does not actually support this feature.  Could call it out as `R410 & BG96-MC not supported` to make it more explicit since those support bands still.

src/content/getting-started/configuration/environment.md
```diff
@@ -0,0 +1,306 @@
...
+
+Sets the preferred list of operators in MCCMNC format. Up to three can be specified. For example:
```
technobly: Up to 4 may be specified.

src/content/getting-started/configuration/environment.md
```diff
@@ -0,0 +1,306 @@
...
+
+- `311480,310410`
+- `311480`
```
technobly, line 222: A note here to say, "In both cases Verizon (311480) will be the first priority.  If you find a device hasn't connected to the preferred operator it may not be available or has missed it in the re-selection process for some reason.  A second attempt can be forced by keeping the same first operator, modifying the subsequent values, and rolling out the changes.  This will force a disconnect and re-selection of the cellular network if the device is not already on the preferred operator."

src/content/getting-started/configuration/environment.md
```diff
@@ -0,0 +1,306 @@
...
+
+- If set to an empty string, the modem default is used
```
technobly: If set to an empty string or variable deleted, the modem default is used

src/content/getting-started/configuration/environment.md
```diff
@@ -0,0 +1,306 @@
...
+
+### PARTICLE_CELLULAR_PREFERRED_BANDS
+
+- Available in Device OS 6.4.0 and later
+- Gen 3 and Gen 4 devices
+- Cellular Modems: BG95-M5, BG96-MC, EG91-E, EG91-EX, EG91-NA, R410, R510, SG560D
+
+Sets preferred bands using a band mask. After 10 minutes of failing to connect, the device reverts to using all bands instead of just the preferred bands.
```
technobly: Please remove: "After 10 minutes of failing to connect, the device reverts to using all bands instead of just the preferred bands." There is no automatic recovery at this time.

technobly (changes requested):

src/content/getting-started/configuration/environment.md
rickkas7: Added, also changed the examples to hex and the calculator tool to generate hex.

src/content/getting-started/configuration/environment.md
rickkas7: Done

src/content/getting-started/configuration/environment.md
rickkas7: Done

src/content/getting-started/configuration/environment.md
rickkas7: Done

src/content/getting-started/configuration/environment.md
rickkas7: Done

src/content/getting-started/configuration/environment.md
rickkas7: Done

src/content/getting-started/configuration/environment.md
rickkas7: Added the text

PR #1581, opened 2026-05-08
Removes old dvos docs

technobly:
@rickkas7 I'm fairly certain I removed everything related to this older DVOS firmware version list, etc.. please feel free to modify if necessary and merge.


rickkas7 (approved):
Thanks! I was just thinking I ought to do this.

PR #1632, opened 2026-08-04
Revert change that removed 802.11a

rickkas7:


rickkas7:
Accidentally pulled in unintended changes, starting over

PR #1640, opened 2026-08-21
Update to Node 24

monkbroc:
Update docs build toolchain to use Node 24.

Verified that the build output is identical in Node 16 and 24.


rickkas7 (approved):
Thanks! I've been meaning to do that.

